PR #89, opened 2020-05-05
Changeset UnitTest TestTpchPlanOnly

TheWanderingMinstrel:
This changeset contains the new test and the statistics data. It has been tested against the unit test suite and is functional.

This changeset contains a unit test (TestTpchPlanONnly) modeled after
TestTpcdsPlanOnly. The statistics file included to run this test is
     tpch/statistics/sf1
and contains statistics data from a 1G run of the TPC-H benchmark.


zhouqingqing:
I will merge this one. Next step is to generate 1g "expected" plan and compare "results" plan during unittest - same as we do for tpch0001.

Thanks,
Qingqing

PR #97, opened 2020-05-19
Changeset with changes for TPC-H 1G Statistical Data

TheWanderingMinstrel:
Qingqing,

This branch contains all the changes needed to run the TPC-H 1G statistics data. The changes are on a branch that is up to date with the master. I have left a few lines of "commented out code" in this changeset because I am not certain how you wanted to handle some matters such as running tests on TPC-H with statistics data from "analyzed table data" vs statistics data read from file. As a result, I think this branch may need a few additional changes to the unit tests before it is merged. Let me know what you think and if you want me to make the changes. 

There is only one query that causes problems when run against the TPC-H 1G data:
     query10 -- using default substitutions
I have skipped this query in the "runFolderandVerify: test and also "if(0)"'ed it out in the TestTpch test where it appears as "File.ReadAllText(files[9]". I don't know if this is a fault of errors in the statistics data or in the plan generated based on this data. 

Finally, I  commented the "try/catch" in the method:
     public static List<Row> ExecSQL
that appears in stmt.cs on line 119 in order to do some debugging of the code I was working on. When I did this, I found some failures in unit tests that were unrelated to my work. I did not investigate these closely, but I did investigate one failure, the unit TestCopy. In this case, the test fails because the filename  specified in the test:
     string filename = @"'..\..\..\data\test.tbl'";  
expands during the test to a wrong directory path. When the "try/catch" are not commented out, this test passes. I suggest that the unit tests should be run without this "try/catch" to verify that the failures that occur are expected and that it is safe to have them ignored and show the tests  as passing. 

Don

--------------------
This changeset contains two files that have the statistics data
for the TPC-H 1G table data set and modifications necessary to
successfully read in this data. It also contains modifications
to the TPC-H unit tests to test reading and using this data.

Finally, this changeset contains the "expected" results files
for the RunFolderAndVerify unit test for TPC-H 1G data.


test/regress/expect/tpch1/q01.txt
```diff
@@ -0,0 +1,37 @@
...
+	l_linestatus
+Total cost: 11882428.35
+PhysicOrder  (inccost=11882428.35, cost=11.35, rows=6) (actual rows=4)
+    Output: l_returnflag[0],l_linestatus[1],{sum(l_quantity)}[2],{sum(l_extendedprice)}[3],{sum(l_extendedprice*1-l_discount)}[4],{sum(l_extendedprice*1-l_discount*1+l_tax)}[5],{avg(l_quantity)}[6],{avg(l_extendedprice)}[7],{avg(l_discount)}[8],{count(*)(0)}[9]
+    Order by: l_returnflag[0], l_linestatus[1]
+    -> PhysicHashAgg  (inccost=11882417, cost=5881202, rows=6) (actual rows=4)
+        Output: {l_returnflag}[0],{l_linestatus}[1],{sum(l_quantity)}[2],{sum(l_extendedprice)}[3],{sum(l_extendedprice*1-l_discount)}[4],{sum(l_extendedprice*1-l_discount*1+l_tax)}[5],{avg(l_quantity)}[6],{avg(l_extendedprice)}[7],{avg(l_discount)}[8],{count(*)(0)}[9]
+        Aggregates: sum(l_quantity[2]), sum(l_extendedprice[3]), sum(l_extendedprice[3]*1-l_discount[7]), sum(l_extendedprice[3]*1-l_discount[7]*1+l_tax[10]), avg(l_quantity[2]), avg(l_extendedprice[3]), avg(l_discount[7]), count(*)(0)
+        Group by: l_returnflag[0], l_linestatus[1]
+        -> PhysicScanTable lineitem (inccost=6001215, cost=6001215, rows=5881190) (actual rows=5914)
```
zhouqingqing, line 30: "actual rows = 5914" - this is not right. 

We shall run against an empty TPCH dataset. More precisely, we shall only do EXPLAIN command with tpch1g stats load without actually running it (but we now don't have EXPLAIN implemented @arzuschen, so a workaround is to run emtpy dataset).

qpmodel/Statis.cs
```diff
@@ -411,12 +411,68 @@ public long EstCardinality(string tabName)
...
+                        return candidate_value;
+
+                case JsonValueKind.Number:
+                    double double_candidate;
+                    //decimal decimal_candidate;
+                    int int_candidate;
+
+                    if (value.TryGetInt32(out int_candidate))
+                        return int_candidate;
+                    //if (value.TryGetDecimal(out decimal_candidate))
```
zhouqingqing, line 437: please remove fake comment

test/UnitTest.cs
```diff
@@ -273,10 +277,12 @@ void TestTpcdsWithData()
         void TestTpchPlanOnly()
         {
             var files = Directory.GetFiles(@"../../../tpch", "*.sql");
-            string stats_fn = "../../../tpch/statistics/sf1";
+            //string stats_fn = "../../../tpch/statistics/sf1";
+            string stats_fn_1 = "../../../tpch/statistics/TPC-H_1G_Statistical_Data/sf1-withoutLineitem";
```
zhouqingqing, line 281: is it possible to manually merge these two?

qpmodel/Statis.cs
TheWanderingMinstrel: done

test/UnitTest.cs
TheWanderingMinstrel: Yes... I am in the process of merging the two statistics files now.

test/regress/expect/tpch1/q01.txt
TheWanderingMinstrel: done

test/UnitTest.cs
TheWanderingMinstrel: To be filed and completed on another branch...

zhouqingqing:
please resubmit

PR #98, opened 2020-05-19
Changeset with changes for TPC-H 1G Statistical Data

TheWanderingMinstrel:
This is a revised version of the TPC-H 1G branch. I  believe it has everything that was requested. If not,, let me know. 

++++++++++

This changeset contains two files that have the statistics data
for the TPC-H 1G table data set and modifications necessary to
successfully read in this data. It also contains modifications
to the TPC-H unit tests to test reading and using this data.

Finally, this changeset contains the "expected" results files
for the RunFolderAndVerify unit test for TPC-H 1G data.


test/UnitTest.cs
```diff
@@ -227,15 +228,9 @@ public void TestBenchmarks()
             TestTpcdsPlanOnly();
             TestTpcdsWithData();
 
+            Tpch.CreateTables();
             TestTpchPlanOnly();
-            TestTpch();
```
zhouqingqing, line 231: why do you remove previous tests? Your tpch1g with 1g real stats against empty tables shall be a new test?

TheWanderingMinstrel: Ok... so there are several things going on here...

**Here's what I did:**

1. I renamed TestTpch to TestTpchWithData. Same test, just renamed for symmetry with TestTpcdsWithData.

2.  TestTpchPlanOnly, is a a mirror image of TestTpcdsPlanOnly with empty tables and 1g real stats. It also contains a setup to run RunFolderAndVerify.  

3. I  took RunFolderAndVerify out of TestBenchmarks because while working 1g, it seemed to be redundant. 


**Here's what I think you wanted:**

1. OK to rename TestTpch to TestTpchWithdata, but not necessary

2. Add TestTpchPlanOnly as a mirror of TestTpcdsPlanOnly and to run empty Tpch tables against some other stats file?

3.  Add TestTpch1g as a mirror image of TestTpcdsPlanOnly to run empty tpch tables against 1g stats and add RunFolderAndVerify to the end of this test to verify plans. 

4. Leave the existing  call  to RunFolderAndVerify at the end of BestBenchmarks. 


Is this correct??

zhouqingqing: too complicated ...

We want one more tests loading 1g stats but running against empty tables. The generated plan shall be verified against "expected" results. Do not remove old tests. Of course, refactoring to avoid code duplication is good.

Is this what implemented?

TheWanderingMinstrel: So as things stand now, TestTpchPlanOnly is the name I have given to TestTpch1g. The plan in this test is generated against 1G stats and verified against expected results. 

The only other difference is that that RunFolderAndVerify is no longer at the end of  TestBenchMarks. 


Do  you want me to rename my new test to TestTpch1g and leave the current TestTpchPlanOnly?

Do you want RunFolderAndVerify called from my new test and from TestBenchmarks?

PR #103, opened 2020-05-22
Migrate to dotnet core

dovahcrow:
This is an initial attempt to migrate the whole project onto dotnet core. 

The project can be successfully run by `dotnet run --project qpmodel` with the following output:

```
select tumble_start(a0, interval '10' second), tumble_end(a0, interval '10' second), count(*) from ast group by tumble(a0, interval '10' second)
***************** raw plan *************
LogicAgg
    Group by: tumble(a0[0],10second)
    -> LogicScanTable ast

***************** optimized plan *************
LogicAgg
    Group by: tumble(a0[0],10second)
    -> LogicScanTable ast


Memo:
0:      1, 1, [42292240][NaN]: ast,PScan(ast: 1)
*1:     1, 1, [-1941974609][NaN]: Agg(0),PHashAgg(0: 3)
---------------------------
Summary: 2,2

Memo:
0:      1, 1, [42292240][1]: ast,PScan(ast: 1)
*1:     1, 1, [-1941974609][4]: Agg(ast),PHashAgg($PScan(ast: 1): 3)
---------------------------
Summary: 2,2
***************** Memo plan *************
Total cost: 4
PhysicHashAgg  (inccost=4, cost=3, rows=1) (actual rows=0)
    Output: {tumble_start(a0,10second)}[1],{tumble_end(a0,10second)}[2],{count(*)(0)}[3]
    Aggregates: tumble_start(a0[0],10second), tumble_end(a0[0],10second), count(*)(0)
    Group by: tumble(a0[0],10second)
    -> PhysicScanTable ast (inccost=1, cost=1, rows=1) (actual rows=0)
        Output: a0[0],10second

-- profiling plan --
Total cost: 4
PhysicHashAgg  (inccost=4, cost=3, rows=1) (actual rows=0)
    Output: {tumble_start(a0,10second)}[1],{tumble_end(a0,10second)}[2],{count(*)(0)}[3]
    Aggregates: tumble_start(a0[0],10second), tumble_end(a0[0],10second), count(*)(0)
    Group by: tumble(a0[0],10second)
    -> PhysicScanTable ast (inccost=1, cost=1, rows=1) (actual rows=0)
        Output: a0[0],10second

RunTime: 00:00:00.0559630
```

However there are several things I want to resolve before this can be merged.

1. Although this project can be successfully run now, it will output some warnings:
> /home/youngw/projects/qpmodel/qpmodel/qpmodel.csproj : warning NU1701: Package 'Microsoft.CodeDom.Providers.DotNetCompilerPlatform 3.5.0-preview1' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v3.1'. This package may not be fully compatible with your project.

I did a search and it says this warning doesn't affect the actual functionalities. I guess we need to somehow suppress this warning in the `csproj` file to make things tidier.

2. A bunch of dependencies are commented out in `qpmodel.csproj`.
    Although they are commented out, it seems like the library just works fine. Maybe we can safely remove these unused dependencies? Or can anyone explain to me what are these commented out dependencies used for?

3. The `app.config` files are not used seemingly. Maybe we can also remove them?
4. For the ANTLR4 package I commented out `<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>` but I'm not sure what's this IncludeAssets for. With this line included the compiler complains cannot find ANTLR.


zhouqingqing:
Weiyun,

Thanks for the patch. 

Project configuration: most likely they are automatically generated with VS studio. Have you run unittest and make sure it passes? Some of your questions might be answered by it. Program.Main() is mostly served as an entrance for debugging specific queries.

dovahcrow:
There are 12 out of 42 tests failed.

qpmodel/Program.cs
```diff
@@ -145,19 +146,19 @@ static void Main(string[] args)
...
-            { 
+            {
                 Tpcds.CreateTables();
                 Tpcds.LoadTables("tiny");
                 Tpcds.AnalyzeTables();
                 // 1, 2,3,7,10,
                 // long time: 4 bad plan
                 // 6: distinct not supported, causing wrong result
-                sql = File.ReadAllText("../../../tpcds/q7.sql");
+                sql = File.ReadAllText("../../../../tpcds/q7.sql");
```
dovahcrow, line 161: The file paths changes due to dotnet core

qpmodel/stmt.cs
```diff
@@ -115,30 +115,22 @@ public virtual List<Row> Exec()
...
-                if (option != null)
-                    stmt.queryOpt_ = option;
-                var result = stmt.Exec();
-                physicplan = "";
-                if (stmt.physicPlan_ != null)
-                    physicplan = stmt.physicPlan_.Explain(option?.explain_);
-                error = "";
-                return result;
-            }
-            catch (Exception e)
```
dovahcrow, line 130: Catch everything and return null silently makes debugging super hard. I'm making it more explicitly but if @zhouqingqing have other concerns I can switch it back.

qpmodel/LogicNode.cs
```diff
@@ -1154,7 +1154,7 @@ public partial class LogicScanTable : LogicGet<BaseTableRef>
 
     public class LogicScanFile : LogicGet<ExternalTableRef>
     {
-        public string FileName() => tabref_.filename_;
+        public string FileName() => tabref_.filename_.Trim();
```
dovahcrow, line 1172: the `filename_ ` can have additional space before and at the end of the string, which is due to the single quotes are being replaced by a space in DataType.cs#L281.

dovahcrow: Idk why windows doesn't have the issue.

test/UnitTest.cs
```diff
@@ -1884,9 +1884,9 @@ public void TumbleWindow()
             TU.ExecuteSQL(sql, "2;2;1", out phyplan);
             sql = "select tumble_start(a0, interval '10' second), tumble_end(a0, interval '10' second), " +
                 "count(*) from ast group by tumble(a0, interval '10' second)";
-            TU.ExecuteSQL(sql, "5/12/2020 7:22:10 AM,5/12/2020 7:22:20 AM,2;" +
```
dovahcrow, line 1887: The time format is an issue. On my side the time format is different from the one in the test.

qpmodel/Program.cs
zhouqingqing: ok

test/UnitTest.cs
dovahcrow: Which part of the code controls the time formatting?

qpmodel/stmt.cs
zhouqingqing: agree try/catch is bad here. We need try/catch to catch "expected" error, say you write "select select 1;", it is an expected syntax error. We do have some tests to verify syntax/algebraic/exec errors (SemanticAnalyzeExeception and SemanticExecutionException).

What's causing problems here are unexpected errors. So a reasonable fix is to check errors in catch block and throw unexpected errors.

qpmodel/LogicNode.cs
zhouqingqing: weird but please make sure windows works with .net core

test/UnitTest.cs
zhouqingqing: It may come from the system environment setting - I am using windows box. One way we can enforce some systems settings at start (say put with Catalog.Init()).

dovahcrow: The issue is fixed by constructing a DateTime instance rather than a string literal.

dovahcrow:
There is another issue happens pervasively: the table name in the physical plan somehow is removed while the expected output preserves it.
For example, with an expected plan like this:
```
PhysicNLJoin   (actual rows=3)
    Output: a.a2[0],a.a3[1],a.a1[2]+b.b2[3]
    Filter: a.a1[2]+b.b3[4]>2
    -> PhysicScanTable a  (actual rows=1)
        Output: a.a2[1],a.a3[2],a.a1[0]
        Filter: a.a1[0]>1
    -> PhysicScanTable b  (actual rows=3)
        Output: b.b2[1],b.b3[2]
```

the assertion failed with:
```
 Assert.AreEqual failed. Expected:<a.a2[0],a.a3[1],a.a1[2]+b.b2[3]>. Actual:<a2[0],a3[1],a1[2]+b2[3]>. 
```

zhouqingqing:
> ```
>  Assert.AreEqual failed. Expected:<a.a2[0],a.a3[1],a.a1[2]+b.b2[3]>. Actual:<a2[0],a3[1],a1[2]+b2[3]>. 
> ```

The behavior is controlled by static ExplainOption.show_tablename_ - it can cause issues as it is global. The reason it is global because I need to use it deep in stack and passing it via global is the easiest way.

The problem might be due to how/the order we set this variable in unittest. You can set running order in "Test Explorer" window  by click "Test" column and make it alphabet sorted to see if the issue went away.

qpmodel/Catalog.cs
```diff
@@ -55,7 +55,8 @@ public class ColumnDef
         public override string ToString() => $"{name_} {type_} [{ordinal_}]";
     }
 
-    public class Distribution {
```
zhouqingqing, line 58: How about split the patch into two PRs? One is for migration itself and the other is for code cleanup. In this way, it is easier to follow the changes.

dovahcrow: Actually I've already split the changes into two commits. You can just look at the first commit to see the actual changes.

I'm also happy to split this one into two PRs if you think that is better.

zhouqingqing: Sure - I am good with two commits as well.

dovahcrow:
Wait for me to investigate why `test/regress/expect/tpch0001/q06.txt` changed.

Issue #48, opened 2020-01-22
enhance EXPLAIN command


zhouqingqing:
simple support is there: look up 

        public ExplainOption explain_ = null;

but we need more supports like:
https://www.postgresql.org/docs/devel/sql-explain.html

PR #107, opened 2020-05-27
Enhanced EXPLAIN command

arzuschen:
This branch is responding to #48 .

Previously, the explain is included in the parsing phase, and the output is muted when EXPLAIN is recognized. However, only the basic physical plan can be extracted using this ExecSQL().

The patch includes two part:
1) force to show cost for plain explain
2) implemented EXPLAIN EXECUTE, which executes the query and then output the full information: the estimated cost and the actual number of rows for each physical step. ANALYZE is a keyword that triggers the parsing as in ANALYZE command instead of as an argument for EXPLAIN, so EXECUTE is used instead.

Additional Explain function is implemented for unittest. This will be further implemented in a separate patch.


qpmodel/Plan.cs
```diff
@@ -107,6 +107,8 @@ public class ExplainOption {
         [ThreadStatic]
         public static bool show_tablename_ = true;
         public bool show_cost_ { get; set; } = false;
+        public bool show_actual_ { get; set; } = true;
```
zhouqingqing, line 110: Can we merge show_cost_ and show_actual_?

qpmodel/Plan.cs
```diff
@@ -159,11 +162,14 @@ public string Explain(ExplainOption option = null, int depth = 0)
                         r += $" (inccost={incCost}, cost={cost}, rows={phynode.logic_.Card()})";
                     }
 
-                    var profile = phynode.profile_;
-                    if (profile.nloops_ == 1 || profile.nloops_==0)
-                        r += $" (actual rows={profile.nrows_})";
-                    else
-                        r += $" (actual rows={profile.nrows_ / profile.nloops_}, loops={profile.nloops_})";
+                    if (exp_showactual)
```
zhouqingqing, line 165: don't duplicate non-trivial logics

qpmodel/SQLParser.cs
```diff
@@ -569,7 +575,14 @@ public override object VisitSql_stmt([NotNull] SQLiteParser.Sql_stmtContext cont
                 throw new NotImplementedException();
 
             Debug.Assert(!r.explainOnly_);
-            r.explainOnly_ = isExplain;
+            if (isExplain)
+            {
+                if (!isExecute)
+                    r.explainOnly_ = true;
```
zhouqingqing, line 581: Can we merge explainOnly_ and explainAnalyze_?

qpmodel/stmt.cs
```diff
@@ -90,12 +92,25 @@ public virtual List<Row> Exec()
...
+            {
+                queryOpt_.explain_.no_output_ = true;
+                queryOpt_.explain_.show_cost_ = true;
+                queryOpt_.explain_.show_actual_ = true;
+            }
 
             // actual execution is needed
             var finalplan = new PhysicCollect(physicPlan_);
-            physicPlan_ = finalplan;
+            //physicPlan_ = finalplan;
```
zhouqingqing, line 113: do not use comment as delete line

qpmodel/stmt.cs
```diff
@@ -123,7 +144,7 @@ public static List<Row> ExecSQL(string sql, out SQLStatement stmt, out string ph
                 var result = stmt.Exec();
                 physicplan = "";
                 if (stmt.physicPlan_ != null)
-                    physicplan = stmt.physicPlan_.Explain(option?.explain_);
+                    physicplan = stmt.physicPlan_.Explain(stmt.queryOpt_.explain_);
```
zhouqingqing, line 147: change option outside?

qpmodel/stmt.cs
```diff
@@ -140,6 +161,29 @@ public static List<Row> ExecSQL(string sql, out SQLStatement stmt, out string ph
         public static List<Row> ExecSQL(string sql, out string physicplan, out string error, QueryOption option = null)
         =>  ExecSQL(sql, out _, out physicplan, out error, option);
 
+        public static void ExplainSQL(string sql, out string physicplan, out string error, QueryOption option = null)
```
zhouqingqing, line 164: do not duplicate non-trivial logics

PR #109, opened 2020-06-03
Patch for statistics part

arzuschen:
This branch deal with the collection of statistics and cardinality estimation. The changes are mainly in three parts, the fourth part is intended but not successful.

1. Implemented CE for OR
Previously, only expressions connected by AND can be recognized by CE process, the expressions are parsed into andlists. Now, orlist is added alongside of andlist, splitting the whole binary expression into tree of andlist and orlist, and the selectivity is estimated accordingly.
For A and B : selectivity = sel(A) * sel(B)
For A or B : selectivity = sel(A) + sel(B) - sel(A)*sel(B)

2. Implemented grouping of same column filters
Under the same level of and/or list, the expressions on the same column is been considered together to produce better result. Only simple expressions of =/</>/<=/>= are included in this process.
For A and B : selectivity = 1 - (1 - sel(A)) - (1 - sel(B))
For A or B : selectivity = sel(A) + sel(B)
This is based on assumption of numerical value and reasonable expressions. For and situation, it is assumed that each expression eliminates some part of the data, and the eliminated parts does not overlap (otherwise the two should be written as one). For or, it is assumed that expression is not overlapping. 

3. Rewrite histogram and MCV
The original column stat is either MCV or histogram. There were some minor issue concerning </<= and >/>= when the cardinality for = is significant.
To improve the estimation accuracy, the method in PostgreSQL is applied. [link](https://www.postgresql.org/docs/12/row-estimation-examples.html)
MCV is always collected except for the unique case, and histogram is also collected even when MCV is collected. Histogram is only considering the part where MCV is not covering. 
1) For = expression, only MCV is used: when the val is in MCV, it is trivial; when not in MCV, the estimation is based on the assumption that all the other values have similar frequency.
2) For inequalities, both stats are used. Scan through MCV to find the sum of frequencies for satisfying values, then use the histogram to find the placement among bins.

(not actual change)
 4. Update join estimation
The method in PostgreSQL is that CE of join is:
card(left) * card(right) / max( distinct(left_col.orig), distinct(right_col.orig) )
As in our previous implementation it was min. Changing this cause infinite recursion problem.

### DIFF
--- a/qpmodel/LogicCard.cs
+++ b/qpmodel/LogicCard.cs
@@ -183,7 +183,7 @@ namespace qpmodel.logic
                     mindlr = 0;
                     break;
                 }
-                mindlr = mindlr * Math.Max(dl, dr);
+                mindlr = mindlr * Math.Min(dl, dr);
             }
             
             if (mindlr != 0)
--- a/qpmodel/LogicCard.cs
+++ b/qpmodel/LogicCard.cs
@@ -183,9 +183,9 @@ namespace qpmodel.logic
                     mindlr = 0;
                     break;
                 }
-                mindlr = mindlr * Math.Min(dl, dr);
+                mindlr = mindlr * Math.Max(dl, dr);
             }
-
+            
             if (mindlr != 0)
                 card = Math.Max(1, (cardl * cardr) / mindlr);
             else
--- a/qpmodel/Statis.cs
+++ b/qpmodel/Statis.cs
@@ -100,11 +100,12 @@ namespace qpmodel.stat
         public double? EstSelectivity(string op, Value val)
         {
             double selectivity = StatConst.one_;
-
+            
             if (!new List<String>() { "=", ">", ">=", "<", "<=" }.Contains(op))
                 return null;
 
             int which = whichBucket(val);
+            
             switch (op)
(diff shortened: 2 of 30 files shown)


qpmodel/Statis.cs
```diff
@@ -102,7 +102,7 @@ int whichBucket(Value val)
             if (val is DateTime)
                 frac = (double)((dval - dlb).Divide(dub - dlb));
             else
-                frac = (double)(dval - dlb) / (dub - dlb);
```
zhouqingqing: > the join estimation have to be changed back to previous due to q95 of
tpcds not working: the node tree have a cycle.

What's that problem？Can you put an issue tracking it?

qpmodel/LogicCard.cs
```diff
@@ -183,9 +183,9 @@ public override ulong LogicJoinCE(LogicJoin node)
                     mindlr = 0;
```
zhouqingqing: > When histogram is not present, force mcv estimation to work on values not in mcv.
In PG, mcv is a complementary to histogram, meaning mcv contains values not presented in histgram. Are you changing to that route?

PR #118, opened 2020-06-11
Changeset implementing SQL Analyze with Sampling

TheWanderingMinstrel:
This changeset has the changes you requested. I took the option changes out of the method AnalyzeStmt. I had originally included them after errors showed up initially and failed to recheck if they were needed after those errors were fixed. Testing now shows that changing the options here is not needed.

TableSample is actually already passed to AnalyzeStmt as part of the parameter: 
     BaseTableRef target
but I added an additional parameter, 
     SelectStmt.TableSample ts
to AnalyzeStmt as requested. I can always remove this new parameter if you think the value should be picked up from "target" instead. 

++++++++++++++++++++++

This changeset allows an SQL analyze statement to gather table
statistics using either a full scan
     Analyze <table name>
 or with table sampling
     Analyze <table name> tablesample row (<num>)
 where "num" is the number of elements in the sample array.

### DIFF
--- a/qpmodel/Statis.cs
+++ b/qpmodel/Statis.cs
@@ -619,8 +619,12 @@ namespace qpmodel.stat
         {
             // its child is a [gather ontop] scan
             var child = child_();
-            Debug.Assert(child is LogicGather || child is LogicScanTable);
-            if (child is LogicGather)
+
+            Debug.Assert(child is LogicGather ||
+                         child is LogicScanTable  ||
+                         child is LogicSampleScan);
+
+            if (child is LogicGather || child is LogicSampleScan)
                 child = child.child_();
             return (child as LogicScanTable).tabref_;
         }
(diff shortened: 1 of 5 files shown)


qpmodel/Statis.cs
```diff
@@ -619,8 +619,12 @@ public BaseTableRef GetTargetTable()
         {
             // its child is a [gather ontop] scan
             var child = child_();
-            Debug.Assert(child is LogicGather || child is LogicScanTable);
-            if (child is LogicGather)
+
```
zhouqingqing, line 622: do you still need to change this method since you don't invoke it anymore?

TheWanderingMinstrel: Apparently so. The method GetTargetTable is still called from PhysicAnalyze.Open.

When the line
> var code = finalplan.Open(context);
is executed for the SQL statement
> analyze a tablesample row (15)
PhysicAnalyze.open calls
> LogicAnalyze.GetTargetTable

The type of the child node at returned by the method child_() in LogicAnalyze.GetTargetTable at this point is "logic.LogicSampleScan" so I extended the Debug.Assert.

Should the processing not follow this path for processing an Analyze statement that contains a tablesample reference?

Issue #122, opened 2020-06-16
CE error for subquery


arzuschen:
The cardinality estimation for join is off when there exists subquery (even the simplest ones)
e.g.
```c#
sql = @"select l_orderkey from (select * from lineitem) l, orders, part
        where l.l_orderkey = o_orderkey and l.l_partkey = p_partkey
        group by l.l_orderkey order by l.l_orderkey";
```
The result contains:
```
    -> PhysicHashAgg 1319_1605  (inccost=341132, cost=300002, rows=1) (actual rows=1500)
        Output: {l_orderkey}[0]
        Group by: l_orderkey[0]
        -> PhysicHashJoin 1428_1607  (inccost=41130, cost=15010, rows=6005) (actual rows=6005)
            Output: l_orderkey[1]
            Filter: l_orderkey[1]=o_orderkey[0]
```
Aggregation node is not getting the correct cardinality estimation even though the join node displays the right number. It should be 6005 rather than 300000 (as indicated by the cost).

PR #124, opened 2020-06-17
Fix inconsistent cost estimation

arzuschen:
This commit targets at #122 

The reason for the inconsistency is that when subquery is present, different join plan may result in different cardinality estimation, current method considers only the default logic plan, and may lead to incorrect cost estimation.
This fix force the cost of a physical node to be updated during the process of finding min cost member, and the cardinality is set as that of the minmember of the group (if availble).

The result of the query mentioned in #122 after this update is
```
    -> PhysicHashAgg 1311_1597  (inccost=47137, cost=6007, rows=1, memory=8) (actual rows=1500)
        Output: {l_orderkey}[0]
        Group by: l_orderkey[0]
        -> PhysicHashJoin 1420_1599  (inccost=41130, cost=15010, rows=6005, memory=12000) (actual rows=6005)
            Output: l_orderkey[1]
            Filter: l_orderkey[1]=o_orderkey[0]
```
Now the CE for the hash join is correctly passed to the cost estimation of hash aggregation.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -140,10 +140,18 @@ namespace qpmodel.physic
         }
 
         #region optimizer
-        public ulong Card() => logic_.Card();
-        public double Cost()
+        public ulong Card()
         {
-            if (double.IsNaN(cost_))
+            if (this is PhysicMemoRef pmr && pmr.Group().minMember_ != null)
+                return pmr.Group().minMember_.physic_.logic_.Card();
+            return logic_.Card();
+        }
+        public double Cost(bool findopt = false)
+        {
+            // force update when finding min cost member
+            if (findopt)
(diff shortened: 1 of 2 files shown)


qpmodel/PhysicalNode.cs
```diff
@@ -140,10 +140,18 @@ public Row ExecProject(Row input)
...
         {
-            if (double.IsNaN(cost_))
+            if (this is PhysicMemoRef pmr && pmr.Group().minMember_ != null)
+                return pmr.Group().minMember_.physic_.logic_.Card();
+            return logic_.Card();
+        }
+        public double Cost(bool findopt = false)
+        {
+            // force update when finding min cost member
+            if (findopt)
```
zhouqingqing, line 152: this is not the right fix: Cost() is cached after it is esimated.

PR #128, opened 2020-06-23
Fix join CE with subquery as join children (#122)

arzuschen:
This commit targets at #122 

Previously, when subquery is present, the cardinality estimation for different joining order may produce different results. This breaks the assumption that the members in the same memogroup should have the same output cardinality, and further lead to inconsistent cost estimation for parent groups.
The last fix (#124 ) focused on finding the cardinality of the min-cost member during cost calculation, instead of making sure that all cardinalities of the same group are the same. 
This fix targets at the cardinality estimation of join. Columns from subquery (FromQueryRef) are referenced back to column map, which contains reference to the original table, then the number of distinct values can in turn be extracted.

Subquery with join operation is also tested, and all unit tests are able to pass.

### DIFF
--- a/qpmodel/LogicCard.cs
+++ b/qpmodel/LogicCard.cs
@@ -157,6 +157,25 @@ namespace qpmodel.logic
         //
         public override ulong LogicJoinCE(LogicJoin node)
         {
+            ulong getDistinct(Expr key)
+            {
+                if (key is ColExpr col)
+                {
+                    var tr = col.tabRef_;
+
+                    if (tr is FromQueryRef fqr && fqr.MapOutputName(col.colName_) != null)
+                        tr = (fqr.MapOutputName(col.colName_) as ColExpr).tabRef_;
+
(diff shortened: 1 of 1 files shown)


qpmodel/LogicCard.cs
```diff
@@ -157,6 +157,25 @@ public override ulong LogicAggCE(LogicAgg node)
...
         public override ulong LogicJoinCE(LogicJoin node)
         {
+            ulong getDistinct(Expr key)
+            {
+                if (key is ColExpr col)
+                {
+                    var tr = col.tabRef_;
+
+                    if (tr is FromQueryRef fqr && fqr.MapOutputName(col.colName_) != null)
+                        if (fqr.MapOutputName(col.colName_) is ColExpr ce) tr = ce.tabRef_;
```
zhouqingqing, line 167: does this fix change any unittests results? if not, I'd like to add one tests under TestJoin test.

Issue #91, opened 2020-05-07
CE and cost model todo items


zhouqingqing:
we have some basic implementation utilizing collected stats in place but need to check what are missing. 

Implemented:
- stats collection with ANALYZE command: we collect min/max, histgram, mcv etc.
- CE of each individual operators (most?)
- CE of simple filters

Todo:
- write primitives in unittest to verify CE and costing correctness
- find out what missing (for example #43 ) and patch them

PR #130, opened 2020-06-23
Cardinality Estimation Unit Tests

arzuschen:
This check-in consists of 2 main parts

1. Cardinality Estimation Unit Tests (#91 ) 
The test is implemented as text comparison between physical plan and the expected plan.
The tests contains 3 categories:
1) filter tests
2) join tests
3) aggregation tests
The test for multiple join with subquery is also for CE of join.

2. Enable string comparison in filer
String comparisons (e.g. ... where some_col < "some_string") are enabled in PostgreSQL. Flexibility is added to statistics (CE) and expression execution.


test/regress/expect/ce/ce0.out
```diff
@@ -0,0 +1,4 @@
+Total cost: 6005
```
zhouqingqing, line 1: is it possible to merge ce0.out, ce0.out ... -> ce.out? Otherwise, we have too many files.

PR #133, opened 2020-06-23
Changeset that adds imdb statistics for jobench suite

TheWanderingMinstrel:
This changeset contains the statistics file for jobench. 

The statistics gathered for the table movie_info_idx are somewhat of a problem. The fourth field of this table is
- named "info"
- described as a text field
- has some very strange values in it
The problem is that these values, which are read strings because they are text, are used with the '<' and '>' operators. The method 

double calcTotalFreq(Value val, string op)" 

cannot  handle this. In addition, it is very difficult to figure how these operators should actually function with the values in this field. As a result, I have moved a number of jobench queries to a "SkippedQueries" subdirectory because they cannot currently be successfully executed. 

I have left the implementation of TestJobench pretty rudimentary for now so that I can research what the values in movie_info_idx represent and how they should be handled. 

--------------------------------------------------------------

This changeset contains the statistics for the jobench suite
tables. It also makes basic modifications to the Unit Test
TestJobench to use the statistics.

This changeset also moves a number of queries to a subdirectory
"skipped Queries" because these queries attempt to compare
strings in a fashion not supported by qpmodel.

Finally, a couple of small fixes were made to the JSON post-
processing in Statis.cs for errors that were not caught in
previous unit tests.


qpmodel/Statis.cs
```diff
@@ -200,7 +200,7 @@ int compare(dynamic v1, dynamic v2)
                 if (which == -1) return otherfreq_;
                 return freqs_[which];
             }
-            
+
             double selectivity = calcTotalFreq(val, op);
```
zhouqingqing, line 204: > double calcTotalFreq(Value val, string op) cannot handle this.

Have you tried lastest master code? It changed calcTotalFreq() to handle string/string comparison.

qpmodel/Statis.cs
```diff
@@ -200,7 +200,7 @@ int compare(dynamic v1, dynamic v2)
                 if (which == -1) return otherfreq_;
                 return freqs_[which];
             }
-            
+
```
zhouqingqing, line 203: do you want to redo the change since the stats '>' already fixed?

TheWanderingMinstrel:
This MR passes the checks that are now automatically run as part of the push process. 

This MR has added the file 

> test/regress/output/jobench/1a.txt

without this file, the unit tests that are run when a branch is pushed will fail. This is because git does not appear to allow an empty "directory" to be pushed. The only way to establish the existence of this directory is to add a file from it. 

It  is  also necessary to override the ignore file in order to add this directory.

test/regress/output/jobench/1a.txt
```diff
@@ -0,0 +1,49 @@
+SELECT MIN(mc.note) AS production_note,
```
zhouqingqing, line 1: You can add a gitkeep file there like this: https://github.com/zhouqingqing/qpmodel/blob/master/test/regress/output/tpch1/.gitkeep

Issue #121, opened 2020-06-16
support bottom select statement with MEMO


zhouqingqing:
By "bottom select" meaning query with bottom half is selection:-
- insert ... select * 
- update ... select *
- copy statement

The reason we don't support them now is because MEMO currently assumes the whole query is a selection.

zhouqingqing:
fixed

PR #134, opened 2020-06-24
Enable memo for insert-select statement (#121)

arzuschen:
Previously all insert statements have use_memo_ disabled. (#121 )
Now insertion with select statement can utilize memo, 
it is done by extracting the select part and use memo only for that.


qpmodel/stmt.cs
```diff
@@ -815,6 +817,9 @@ internal void OpenSubQueries(ExecContext context)
                 v.query_.OpenSubQueries(context);
             context.option_.PopCodeGen();
         }
+
+        public override SelectStmt ExtractSelect() => this;
+        public override PhysicNode MemoOpt(PhysicNode select) => select;
```
zhouqingqing, line 822: rename MemoOpt => InstallSelectPlan?

arzuschen: Sure, I will change that.

PR #146, opened 2020-07-03
property enforcement in memo optimization

arzuschen:
Established the property enforcement process in optimization.
A property propagation step is created between exploration and cost calculation. In this step, properties are propagated from top to bottom, creating link between required properties and the group members that can supply them. Then the cost calculation determine the optimal case for each group under each required property. 
However, only exact match of the supplied property is considered in the model and partial property propagation is not implemented. This can be done through modification of the property method.
Properties are enforced on top of the original member (instead of down below in the previous implementation), the additional physical node will be the parent of the existing member that does not supply the property. This enables base requirement to be added.

Logic nodes on the top like ORDER is transformed into physical property to prevent corresponding groups to be create. For example, if an order node is on the top of the raw plan, it will be added to the base required property rather than a memo group. The new root group would be the first logic node that cannot be transformed into physical property.

All unittest pass.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -65,7 +65,7 @@ namespace qpmodel.physic
         public override string ExplainOutput(int depth, ExplainOption option) => logic_.ExplainOutput(depth, option);
         public override string ExplainInlineDetails() => logic_.ExplainInlineDetails();
         public override string ExplainMoreDetails(int depth, ExplainOption option) => logic_.ExplainMoreDetails(depth, option);
-
+        
         public void ValidateThis()
         {
             VisitEach(x =>
@@ -161,7 +161,7 @@ namespace qpmodel.physic
             children_.ForEach(x =>
             {
                 if (x is PhysicMemoRef xp)
-                    incCost += xp.Group().minIncCost_;
+                    incCost += xp.Group().nullMinIncCost_;
                 else
                     incCost += x.InclusiveCost();
             });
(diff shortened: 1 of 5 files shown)


qpmodel/PhysicalNode.cs
```diff
@@ -161,7 +161,7 @@ public double InclusiveCost()
             children_.ForEach(x =>
             {
                 if (x is PhysicMemoRef xp)
-                    incCost += xp.Group().minIncCost_;
+                    incCost += xp.Group().nullMinIncCost_;
```
zhouqingqing, line 164: better split refactoring code (no or trivial semantic changes, but with rename/function refactoring work) and actual working code.

qpmodel/optimizer.cs
```diff
@@ -44,10 +45,80 @@
...
+    {
+        KeyValuePair<ColExpr, string> distributed_;
+        List<Expr> ordered_ = new List<Expr>();
+        List<bool> descends_ = new List<bool>();
+
+        public PhysicProperty() { }
+        public PhysicProperty(List<Expr> exprlist, List<bool> desclist = null)
+        {
+            foreach (Expr expr in exprlist)
+                ordered_.Add(expr);
```
zhouqingqing, line 69: => ordered_.AddRange(exprlist);

qpmodel/optimizer.cs
```diff
@@ -292,54 +367,144 @@ public void CountMembers(out int clogic, out int cphysic)
...
+        }
+
+        public CGroupMember locateMinCostMember(PhysicProperty property)
+        {
+            var nullmin = locateMinCostMember();
+            if (property is null)
+                return nullmin;
+            else
+            {
+                if (propMinMember_.ContainsKey(property.GetHashCode()))
```
zhouqingqing, line 485: hashCode is a signature of an object, two objects may have the same signature. So you can't use hashCode to uniquely identify an object.

PR #151, opened 2020-07-07
Add property propagation in memo optimization

arzuschen:
This patch attempts to add a new property propagation step to the end the exploration. The groups are traversed from top to down and properties required by certain physical nodes are propagated to the child group. If a member in the child group can provided the property, it will be added to the candidate list.

The locate min cost member functions are changed to accommodate to property requirements. If the property requirement is null, then the min cost member is found and returned same as before. When a property requirement is present, the min cost member with additional enforcement node will be compared to all the other members that can provide that property. Then the optimal member will be found.

Since enforced node will be added onto the member physical node, the assumption that each member correspond to one layer of physical node is no longer valid. However, since bypass is created for join solver case, which will also work for enforced node case, minimal changes is required for copy out min logic physic plan function.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -1062,11 +1062,15 @@ namespace qpmodel.physic
         public override string ToString() => $"PStreamAgg({child_()}: {Cost()})";
         public override PhysicProperty RequiredProperty()
         {
-            return new PhysicProperty((logic_ as LogicAgg).groupby_);
+            var exprlist = (logic_ as LogicAgg).groupby_;
+            if (exprlist is null) return null;
+            return new PhysicProperty(exprlist);
         }
         public override PhysicProperty SuppiedProperty()
         {
-            return new PhysicProperty((logic_ as LogicAgg).groupby_);
+            var exprlist = (logic_ as LogicAgg).groupby_;
+            if (exprlist is null) return null;
+            return new PhysicProperty(exprlist);
         }
         protected override double EstimateCost()
         {
@@ -1109,7 +1113,7 @@ namespace qpmodel.physic
                     {
                         // output current grouped row if any
                         if (curGroupKey != null)
-                            FinalizeAGroupRow(context, keys, curGroupRow, callback);
+                            FinalizeAGroupRow(context, curGroupKey, curGroupRow, callback);
 
                         // start a new grouped row
                         curGroupRow = AggrCoreToRow(l);
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -190,6 +190,9 @@ namespace qpmodel.physic
 
             return memory;
         }
+        public virtual PhysicProperty RequiredProperty() => null;
+        public virtual PhysicProperty SuppiedProperty() => null;
+        public virtual PhysicProperty PropagatedProperty() => null;
         #endregion
 
         public BitVector tableContained_ { get => logic_.tableContained_; }
@@ -1057,7 +1060,14 @@ namespace qpmodel.physic
     {
         public PhysicStreamAgg(LogicAgg logic, PhysicNode l) : base(logic, l) { }
         public override string ToString() => $"PStreamAgg({child_()}: {Cost()})";
-
+        public override PhysicProperty RequiredProperty()
+        {
+            return new PhysicProperty((logic_ as LogicAgg).groupby_);
+        }
+        public override PhysicProperty SuppiedProperty()
+        {
+            return new PhysicProperty((logic_ as LogicAgg).groupby_);
+        }
         protected override double EstimateCost()
         {
             return logic_.Card() * 2.0;
--- a/qpmodel/optimizer.cs
+++ b/qpmodel/optimizer.cs
@@ -50,8 +50,44 @@ namespace qpmodel.optimizer
     public class PhysicProperty : Property
     {
         // ordering: the ordered expression and whether is descending
-        KeyValuePair<Expr, bool> ordering_;
-        KeyValuePair<Expr, DistributionType> distribution_;
+        List<KeyValuePair<Expr, bool>> ordering_;
+        List<KeyValuePair<Expr, DistributionType>> distribution_;
+        public PhysicProperty(List<Expr> order, List<bool> desc = null)
+        {
+            if (desc is null)
+                desc = new List<bool>(Enumerable.Repeat(true, order.Count).ToArray());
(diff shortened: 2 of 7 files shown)


qpmodel/optimizer.cs
```diff
@@ -50,8 +51,67 @@ public class Property { }
     public class PhysicProperty : Property
     {
         // ordering: the ordered expression and whether is descending
-        KeyValuePair<Expr, bool> ordering_;
-        KeyValuePair<Expr, DistributionType> distribution_;
+        List<KeyValuePair<Expr, bool>> ordering_;
+        List<KeyValuePair<Expr, DistributionType>> distribution_;
+        public PhysicProperty(List<Expr> order, List<bool> desc = null)
```
zhouqingqing, line 56: https://coveralls.io/builds/31922927/source?filename=qpmodel/optimizer.cs

Seems these code are not exercised in unittest, can you add them?

arzuschen: Sure

zhouqingqing: also, consider removing distribution_ and introducing it in your next PR with unittest.

qpmodel/PhysicalNode.cs
```diff
@@ -1057,7 +1060,18 @@ public class PhysicStreamAgg : PhysicAgg
...
+        {
+            var exprlist = (logic_ as LogicAgg).groupby_;
+            if (exprlist is null) return null;
+            return new PhysicProperty(exprlist);
+        }
+        public override PhysicProperty SuppiedProperty()
+        {
+            var exprlist = (logic_ as LogicAgg).groupby_;
+            if (exprlist is null) return null;
+            return new PhysicProperty(exprlist);
```
zhouqingqing, line 1073: do you mean it supplies (after streamAgg) sorted attributes on group by keys? If so, return "PhysicProperty" is not clear in naming. You may want to derive a "SortOrderProperty" out of Property class for this purpose?

qpmodel/optimizer.cs
```diff
@@ -35,6 +35,7 @@
 using qpmodel.expr;
 
 using LogicSignature = System.Int64;
+using qpmodel.utils;
```
zhouqingqing, line 38: move to L36

qpmodel/optimizer.cs
```diff
@@ -50,8 +51,61 @@ public class Property { }
     public class PhysicProperty : Property
```
zhouqingqing, line 51: See previous comment, you may want to have SortOrderProperty/DistributionProperty etc derived from Property? For clear naming.

qpmodel/optimizer.cs
```diff
@@ -50,8 +51,61 @@ public class Property { }
...
     {
         // ordering: the ordered expression and whether is descending
-        KeyValuePair<Expr, bool> ordering_;
-        KeyValuePair<Expr, DistributionType> distribution_;
+        List<KeyValuePair<Expr, bool>> ordering_;
```
zhouqingqing, line 54: We may have COLLATE in future in sort clause, so sort is ORDER BY <expr> <desc?> <collation?>. You can have a refactoring patch to introduce a SortByExpr class grouping them together.

qpmodel/optimizer.cs
```diff
@@ -50,8 +51,61 @@ public class Property { }
...
+            for (int i = 0; i < ordering_.Count; i++)
+            {
+                if (!ordering_[i].Key.Equals(other.ordering_[i].Key)) return false;
+                if (ordering_[i].Value != other.ordering_[i].Value) return false;
+            }
+            return true;
+        }
+        public override bool Equals(object obj)
+        {
+            if (obj.GetType() != this.GetType()) return false;
```
zhouqingqing, line 102: if (!(obj is PhysicProperty)) return false?

qpmodel/optimizer.cs
```diff
@@ -50,8 +51,61 @@ public class Property { }
...
+                order.Add(pair.Key);
+                desc.Add(pair.Value);
+            }
+            var logicnode = new LogicOrder(node.logic_, order, desc);
+            return new PhysicOrder(logicnode, node);
+        }
+
+        public bool Equals(PhysicProperty other)
+        {
+            if (other is null) return false;
```
zhouqingqing, line 91: no null checks here since you already did it in caller.

qpmodel/optimizer.cs
arzuschen: I wanted to declare that if if is compared to a object that is not PhysicProperty then they cannot be equal.

qpmodel/PhysicalNode.cs
```diff
@@ -1099,7 +1113,7 @@ public override string Exec(Func<Row, string> callback)
                     {
                         // output current grouped row if any
                         if (curGroupKey != null)
-                            FinalizeAGroupRow(context, keys, curGroupRow, callback);
+                            FinalizeAGroupRow(context, curGroupKey, curGroupRow, callback);
```
zhouqingqing, line 1116: please send in a separate PR for this bug fix - I will do a quick merge.

qpmodel/optimizer.cs
```diff
@@ -345,11 +405,44 @@ public void ExploreGroup(Memo memo)
...
+                propertyCandidates_.Add(property, new List<int>());
+            if (property is null)
+            {
+                if (propagated_) return;
+                propagated_ = true;
+            }
+
+            for (int i = 0; i < exprList_.Count; i++)
+            {
+                if (exprList_[i].physic_ != null && !solveroptimized)
```
zhouqingqing, line 427: is it possible to follow ExplorGroup() style like:
  EnforceGroup()
     foreach (member in exprList_)
         EnforceMember();

qpmodel/optimizer.cs
```diff
@@ -713,7 +882,7 @@ public PhysicNode CopyOutOnePlan(SQLStatement stmt, Memo memo)
 
             // retrieve the lowest cost plan
             Debug.Assert(stmt.physicPlan_ is null);
-            stmt.physicPlan_ = memo.rootgroup_.CopyOutMinLogicPhysicPlan();
+            stmt.physicPlan_ = memo.rootgroup_.CopyOutMinLogicPhysicPlan(memo.baseproperty_);
             stmt.logicPlan_ = stmt.physicPlan_.logic_;
```
zhouqingqing, line 887: please print memo for better illustration

qpmodel/optimizer.cs
arzuschen: I think I still need this because the caller only checks for type

PR #156, opened 2020-07-13
add streamagg codegen support

arzuschen:
This patch completed codegen for stream aggregate and resolved the conflict between limit and stream aggregate.
Since hash aggregate only have one result outlet (FinalizeAGroupRow), a generic stop can work. However, stream aggregate has two result outlet, the second outlet may conflict with limit node because stop is not triggered.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -1092,6 +1092,9 @@ namespace qpmodel.physic
                     buildcode += $@"
                         var keys = KeyList.ComputeKeys(context, {_logic_}.groupby_, {lrow});";
                     buildcode += $@"
+                        if (context.stop_)
+                            return;
+
                         if (curGroupKey{_} != null && keys.Equals(curGroupKey{_}))
                         {{
                             for (int i = 0; i < {aggrcore.Count}; i++)
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -1099,7 +1099,7 @@ namespace qpmodel.physic
                     {
                         // output current grouped row if any
                         if (curGroupKey != null)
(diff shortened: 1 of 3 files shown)


qpmodel/PhysicalNode.cs
```diff
@@ -1113,15 +1143,36 @@ public override string Exec(Func<Row, string> callback)
...
+            else
             {
                 if (logic.groupby_ is null && curGroupRow is null)
                 {
                     Row row = HandleEmptyResult(context);
                     if (row != null)
                         callback(row);
                 }
-                if (curGroupKey != null)
+                if (curGroupKey != null && !context.stop_)
```
arzuschen, line 1175: I am having some concern about this change. It does not seem as clean for stop to appear in aggregate node.

zhouqingqing: We may want to stop current execution for some reasons: 1) LIMIT gets enough data; 2) query cancel due to user or abort due to internal error. context.stop_ is designed for the 1st one. The second one will be handled by try/catch exception.

Usually we check stop_ flag on top of each loop thus we won't waste too much before stop.

qpmodel/PhysicalNode.cs
```diff
@@ -1099,7 +1129,7 @@ public override string Exec(Func<Row, string> callback)
                     {
                         // output current grouped row if any
                         if (curGroupKey != null)
-                            FinalizeAGroupRow(context, keys, curGroupRow, callback);
+                            FinalizeAGroupRow(context, curGroupKey, curGroupRow, callback);
```
zhouqingqing, line 1132: I thought this is fixed a while back?

arzuschen: yes I used the same branch, I don't know why this change still appears.

PR #158, opened 2020-07-15
Add order property enforcement in memo optimization

arzuschen:
Migrated from #151 and cleaned, with additional unit test.

The three main parts are property propagation, property enforcement and property conversion. The property used here only refers to order property, distribution can be added to this framework.
Prior to exploration phase, top logic nodes that can be converted into property requirement is added into base requirement of the corresponding memo.
After exploration, a separate property propagation phase is added to find the candidate members of certain property in a memo group.
During the cost computation, property enforcement is added if there is a required property.

Unit test is added to cover the added code, with the exception of leaf group candidate. Since there is no current leaf physic node that supply order property.

PR #159, opened 2020-07-16
Add order property enforcement in memo optimization

arzuschen:
Updated from #158, complexity of implementation is greatly reduced.

Property enforcement is added without property propagation step. The cost calculation step is rewritten into a generalized function, without the separation between leaf group (and solver optimized group) and non-leaf group. Property requirement is added into the calculation of cost.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -168,7 +168,7 @@ namespace qpmodel.physic
                     incCost += x.InclusiveCost();
             });
 
-            Debug.Assert(Double.IsNaN(incCost) || incCost > Cost() || children_.Count == 0);
+            Debug.Assert(incCost > Cost() || children_.Count == 0);
             return incCost;
         }
 
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -192,7 +192,8 @@ namespace qpmodel.physic
         }
         public virtual PhysicProperty RequiredProperty() => null;
         public virtual PhysicProperty SuppiedProperty() => null;
-        public virtual PhysicProperty PropagatedProperty() => null;
+        public virtual List<PhysicProperty> PropagatedProperty(PhysicProperty property) 
+            => new List<PhysicProperty>( new PhysicProperty[children_.Count] );
         #endregion
 
         public BitVector tableContained_ { get => logic_.tableContained_; }
@@ -624,6 +625,17 @@ namespace qpmodel.physic
         {
             double cost = (l_().Card() + 10) * (r_().Card() + 10);
             return cost;
+        }
+
+        public override List<PhysicProperty> PropagatedProperty(PhysicProperty property)
+        {
+            var logic = logic_ as LogicJoin;
+            logic.CreateKeyList(false); // ensure existence of left/right keys
+            
+            if (property != null && property is SortOrderProperty sort)
+                if (sort.IsExprMatch(logic.leftKeys_))
+                    return new List<PhysicProperty> { sort, null };
+            return base.PropagatedProperty(property);
         }
     }
 
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -168,7 +168,7 @@ namespace qpmodel.physic
                     incCost += x.InclusiveCost();
             });
 
-            Debug.Assert(incCost > Cost() || children_.Count == 0);
+            Debug.Assert(Double.IsNaN(incCost) || incCost > Cost() || children_.Count == 0);
             return incCost;
         }
 
@@ -190,6 +190,9 @@ namespace qpmodel.physic
 
             return memory;
         }
+        public virtual PhysicProperty RequiredProperty() => null;
+        public virtual PhysicProperty SuppiedProperty() => null;
+        public virtual PhysicProperty PropagatedProperty() => null;
         #endregion
 
         public BitVector tableContained_ { get => logic_.tableContained_; }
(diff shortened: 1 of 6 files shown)


test/UnitTest.cs
```diff
@@ -653,6 +653,47 @@ public void TestMemo()
...
+            memo.CalcStats(out int tlogics, out int tphysics);
+            Assert.AreEqual(8, memo.cgroups_.Count);
+            Assert.AreEqual(16, tlogics); Assert.AreEqual(17, tphysics);
+            Assert.AreEqual("4,1;6,4", string.Join(";", result));
+            var mstr = stmt.optimizer_.PrintMemo();
+            Assert.IsTrue(mstr.Contains("Summary: 16,17"));
+
+            sql = "select a1 from a, b where a1 <= b1 group by a1 order by a1";
+            result = TU.ExecuteSQL(sql, out stmt, out phyplan, option);
+            memo = stmt.optimizer_.memoset_[0];
```
zhouqingqing, line 686: Can you find a way to assert we can push sort under NLJ?

arzuschen: that should be doable

test/UnitTest.cs
```diff
@@ -119,6 +119,17 @@ public static void CountOccurrences(string text, string pattern, int expected)
             var count = CountStr(text, pattern);
             Assert.AreEqual(expected, count);
         }
+        public static bool OccurenceWithOrder(string text, List<string> patternlist)
```
zhouqingqing, line 122: we shall assert that the order is underneath join? try to check the plan (not the plan string) directly. - you can extend the ExecSQL() interface if we can't get the plan.

qpmodel/PhysicalNode.cs
```diff
@@ -190,6 +190,10 @@ public ulong InclusiveMemory()
 
             return memory;
         }
+        public virtual PhysicProperty RequiredProperty() => null;
```
arzuschen, line 193: add comment here

qpmodel/Program.cs
```diff
@@ -189,6 +189,7 @@ static void Main(string[] args)
             a.queryOpt_.explain_.show_id_ = true;
             a.queryOpt_.explain_.show_estCost_ = a.queryOpt_.optimize_.use_memo_;
             a.queryOpt_.explain_.mode_ = ExplainMode.full;
+            a.queryOpt_.optimize_.enable_streamagg_ = false;
```
arzuschen, line 192: remove this change

qpmodel/optimizer.cs
```diff
@@ -50,10 +50,77 @@ public class Property { }
...
-        KeyValuePair<Expr, DistributionType> distribution_;
+        public List<KeyValuePair<Expr, bool>> ordering_ = new List<KeyValuePair<Expr, bool>>();
+        
+        public bool IsPropertySupplied(PhysicNode node)
+        {
+            if (this.Equals(node.SuppiedProperty()))
+                return true;
+            return false;
+        }
+        // check if the left key of NLJoin match the property
```
arzuschen, line 61: move this part to nljoin

qpmodel/optimizer.cs
```diff
@@ -296,6 +368,17 @@ public string Print()
             CountMembers(out int clogics, out int cphysics);
             var str = $"{clogics}, {cphysics}, [{logicSign_}][{minIncCost_}]: ";
             str += string.Join(",", exprList_);
+
+            // add property optimal member
+            if (minMember_.Count > 0)
+            {
+                List<string> l = new List<string>();
+                foreach (var pair in minMember_)
+                    l.Add($"property:{pair.Key.ToString()}, member:{pair.Value.Key}, cost:{pair.Value.Value.ToString("0.##")}");
```
arzuschen, line 377: remove the first ToString (unnecessary)

qpmodel/optimizer.cs
```diff
@@ -342,102 +425,135 @@ public void ExploreGroup(Memo memo)
             explored_ = true;
         }
 
-        // propagate the property requirement from the top
-        public void PropagateProperty(PhysicProperty property)
+        // preprocess exprList_ to get whether property can be supplied and subproperties
+        // also find members that can propagate required property and split them
```
arzuschen, line 429: add more comments about the input output lists

qpmodel/optimizer.cs
```diff
@@ -342,102 +425,135 @@ public void ExploreGroup(Memo memo)
...
+                    fulfilled.Add(property?.IsPropertySupplied(member.physic_) ?? true);
+                    subproplist.Add(subproperties);
                 }
             }
-
-            minIncCost_ = mincost;
-            Debug.Assert(IsSolverOptimizedGroup() ||
-                minMember_.physic_.InclusiveCost() == minIncCost_);
-            return minMember_;
+            Debug.Assert(candidates.Count == fulfilled.Count && candidates.Count == subproplist.Count);
```
arzuschen, line 458: add assertion here:
if property is null, candidate number = physic member number

qpmodel/optimizer.cs
arzuschen: change to tuple

qpmodel/optimizer.cs
```diff
@@ -342,102 +425,135 @@ public void ExploreGroup(Memo memo)
...
-            double mincost = double.MaxValue;
-            foreach (var v in exprList_)
+            foreach (var member in exprList_)
             {
-                var physic = v.physic_;
-                if (physic != null && physic.Cost() < mincost)
+                if (member.physic_ is null) continue;
+
+                var subproperties = ProcessChildProperty(member.physic_, property, out bool propagated);
+                if (propagated)
```
arzuschen, line 442: add example here

qpmodel/optimizer.cs
```diff
@@ -342,102 +425,135 @@ public void ExploreGroup(Memo memo)
...
+                // when no required property and propagated is not all null
+                // meaning certain property is propagated to child
+                // the status is recorded
+                foreach (var prop in childprop)
+                    if (prop != null) propagated = true;
+            }
+            return childprop;
+        }
+
+        internal CGroupMember CalculateMinInclusiveCostMember(PhysicProperty property = null)
```
zhouqingqing, line 484: let's try to follow the original code path per discussion.

PR #161, opened 2020-07-21
property enforcement with physical order 

arzuschen:
Updated from #159, some minor modifications are made.

The major change is making the property enforcement part to follow the original logic of finding the minimum cost member. The original way was member being looped through and inclusive cost for physical member is computed by summing its cost with the cost of all the members. The minimum cost and the corresponding member is recorded for each group. This enabled copy out tree to acquire the minimum cost member from top to down.
However, with property requirement added into the process, the property required on this group and the property required on the child groups need to be taken into consideration. Meaning that a single cost and member is no long sufficient. A dictionary is attached to the group, recording the different cost/member for each property requirement.
The calculation is trivial when property requirement on current group is null, we simply look for the minimum inclusive cost member. Thus, here we focus on not null property requirement.

When there are required property, there are three mutually exclusive cases for any member:
1) property directly supplied by the member
2) property is able to propagate to child
3) property is neither supplied nor propagated

Another attribute of the member is whether there is property required directly by the member. We assume that if the member has property requirement, then it cannot have the ability to propagate property.
With this in consideration, there are two situation where some property is required on children of the member:
property is propagated or property is required by the member.

Therefore, two (at most) costs are computed for each member of the group:
Cost0: no property on children
Cost1: property requirement on children

Also, two lists are maintained for the group:
List0: any member (no property requirement)
List1: members that the required property can be supplied (either through propagation or direct supply)

Combining property supply and property required by this current member, there are a total of 5 cases:
1.1) property directly supplied by member/ no property required by member--> add cost0 to both list0 and list1
1.2) property directly supplied by member/ there is property required by member--> add cost1 to both list0 and list1
2) property able to propagate to child--> add cost0 to list 0 and add cost1 to list1
3.1) property is neither supplied nor propagated/ no property required by member--> add cost0 to list0
3.2) property is neither supplied nor propagated/ there is property required by member--> add cost1 to list0
These 5 cases can serve as a partition for all the members, mutually exclusive while covering.

After members are looped through to get list0 and list1, find the min inclusive cost member for different property requirement and record in the dictionary. For no requirement, simply find the minimum cost in list0. For some property requirement, take the no requirement minimum, enforce the property and use it as a starting point. Loop through list1 and compare to get the minimum cost/member that satisfy the required property. Also record this in the dictionary.

The property requirement for children is recorded as an attribute in each member. This is used when copying out the physical plan. With this information, we know which member from the dictionary should be used.

### DIFF
--- a/qpmodel/optimizer.cs
+++ b/qpmodel/optimizer.cs
@@ -595,7 +595,6 @@ namespace qpmodel.optimizer
                 // calculate the min member for property requirement
                 if (required != null)
                     GetPropertyMinCostTuple(required, supplied);
-
             }
             Debug.Assert(!double.IsNaN(minIncCost_));
 
--- a/qpmodel/optimizer.cs
+++ b/qpmodel/optimizer.cs
@@ -571,11 +571,14 @@ namespace qpmodel.optimizer
             minMember_.Add(required, new Tuple<CGroupMember, double>(minmember, mincost));
         }
 
-        public CGroupMember CalculateMinInclusiveCostMember(PhysicProperty required = null)
+        public CGroupMember CalculateMinInclusiveCostMember(PhysicProperty required = null, PhysicNode parent = null)
         {
             // inclusive cost is only possible after exploration done
             Debug.Assert(explored_);
             Debug.Assert(exprList_.Count >= 2);
+
+            // if there is parent node requirement, independently do this calculation
+            if (parent != null) CalculateMinInclusiveCostMember(parent.RequiredProperty());
 
             if (required == null && minMember_.ContainsKey(new PhysicProperty()))
                 return minMember_[new PhysicProperty()].Item1;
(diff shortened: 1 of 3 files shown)


qpmodel/optimizer.cs
```diff
@@ -349,95 +431,188 @@ public void PropagateProperty(PhysicProperty property)
...
             }
 
-            minIncCost_ = mincost;
+            // record the null property minimum inclusive cost
+            if(mincost > 0) minIncCost_ = mincost;
+
             Debug.Assert(IsSolverOptimizedGroup() ||
-                minMember_.physic_.InclusiveCost() == minIncCost_);
-            return minMember_;
+                minMember_[new PhysicProperty()].Item1.physic_.InclusiveCost() == minIncCost_);
```
zhouqingqing, line 474: remove return

qpmodel/optimizer.cs
```diff
@@ -349,95 +431,188 @@ public void PropagateProperty(PhysicProperty property)
...
+            return minMember_[required ?? new PhysicProperty()].Item1;
         }
 
-        public CGroupMember CalculateMinInclusiveCostMember()
+        internal void CalculateMemberCosts(PhysicProperty required, CGroupMember member,
```
zhouqingqing, line 478: header comment

zhouqingqing: let's try to pull up this logic with the parent (as it recursively called parent) and simplify with 3=>2 logics indicated in your checkin comment.

PR #162, opened 2020-07-22
Fix property enforcement cost calculation

arzuschen:
1) Modify minIncCost of group:
    Change its name to nullPropertyMinIncCost to represent the real meaning, and only initialize the get method of that member.
2) Fix cost calculation error 
    This error was caused by member referring to the same object, so later changes in the object affect the used object. Problem solved by creating new object for member associated with cost1.
3) Modify unittest
    Adjust one of the property enforcement unittest to force the optimizer push order node below PNLJoin.
4) Change null property
    Create a static member in PhysicProperty to represent null property. This avoid multiple call of default property constructor.

### DIFF
--- a/qpmodel/optimizer.cs
+++ b/qpmodel/optimizer.cs
@@ -595,7 +595,6 @@ namespace qpmodel.optimizer
                 // calculate the min member for property requirement
                 if (required != null)
                     GetPropertyMinCostTuple(required, supplied);
-
             }
             Debug.Assert(!double.IsNaN(minIncCost_));
 
--- a/qpmodel/optimizer.cs
+++ b/qpmodel/optimizer.cs
@@ -571,11 +571,14 @@ namespace qpmodel.optimizer
             minMember_.Add(required, new Tuple<CGroupMember, double>(minmember, mincost));
         }
 
-        public CGroupMember CalculateMinInclusiveCostMember(PhysicProperty required = null)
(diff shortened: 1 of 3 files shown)


qpmodel/optimizer.cs
```diff
@@ -303,7 +303,6 @@ public class CMemoGroup
         //
         public Dictionary<PhysicProperty, (CGroupMember member, double cost)> minMember_ { get; set; }
             = new Dictionary<PhysicProperty, (CGroupMember, double)>();
-        public double minIncCost_ { get; set; }
```
zhouqingqing, line 306: make minIncCostNullProperty_ a geet only property.

zhouqingqing: class NullProperty: PhysicProperty{} so when we enforce property, it is never null.

qpmodel/optimizer.cs
```diff
@@ -597,7 +590,7 @@ public CGroupMember CalculateMinInclusiveCostMember(PhysicProperty required = nu
 
             // check if we already done the calculation
             if (required == null && minMember_.ContainsKey(new PhysicProperty()))
```
zhouqingqing, line 592: if we make required default value to PhysicProperty.nullprop (so never nul) , will this simplify this logics?

arzuschen: the default value for optional parameter cannot be PhysicProperty.nullprop since it has to be compile time constant, I used overload as a bypass

PR #163, opened 2020-07-26
between clause issue#4

korenhe:
I tested these two patches through the test - Unit Tests
which has 44 test cases and all passed.

[1/2] modify the g4 and parser.cs
[2/2] modify all between () -> between .. and in the tests.

In fact I don't have time to read the ANTLR book, and I am not quite sure if this modification is the right/best way, for example , 

- should I separate expr into logical_expr, bool_expr and arith_expr, I don't have theory support but just lots of tries. 
- I use three ( .. ) in logical_expr, arith_expr, and expr, I am sure the expr one should never entered and I will leave an assertion there, when I figure out how to modify this:        
public override object VisitOtherbrackexpr([NotNull] SQLiteParser.OtherbrackexprContext context)
            => Visit(context.expr());

-.-|||

### DIFF
--- a/qpmodel/SQLite.g4
+++ b/qpmodel/SQLite.g4
@@ -162,34 +162,62 @@ column_constraint
     AND
     OR
 */
-expr
- : literal_value											#LiteralExpr
- | BIND_PARAMETER											#bindexpr
- | ( ( database_name '.' )? table_name '.' )? column_name	#ColExpr
- | unary_operator expr										#unaryexpr
- | expr '||' expr											#strconexpr
- | expr op=( '*' | '/' | '%' ) expr							#arithtimesexpr
- | expr op=( PLUS | MINUS ) expr							#arithplusexpr
- | expr op=( '<<' | '>>' | '&' | '|' ) expr					#arithbitexpr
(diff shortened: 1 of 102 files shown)


qpmodel/SQLite.g4
```diff
@@ -162,34 +162,62 @@ column_constraint
...
- | expr K_NOT? K_IN ( '(' ( select_stmt						
-                          | expr ( ',' expr )*
+logical_expr
+ : logical_expr K_AND logical_expr											#LogicAndExpr
+ | logical_expr K_OR logical_expr											#LogicOrExpr
+ | bool_expr                                                         #boolexpr
+ | '(' logical_expr ')'												#brackexpr
+;
+
+bool_expr
```
zhouqingqing, line 172: rename as predicate_expr? because bool_expr shall also contain logic_expr?

zhouqingqing: > should I separate expr into logical_expr, bool_expr and arith_expr, I don't have theory support but just lots of tries.

I don't think there is a definite rule how they shall be splitted. Current one looks good to me.

zhouqingqing: what is three ( .. )?

PR #164, opened 2020-07-27
Support ANTLR4 and auto generation csharp files

korenhe:
The csharp support for ANTLR4 is still experimental and need to be enabled
manually with some properties in csproj file.

This patch prevent the occasionally wrong generated java files in obj folder.


zhouqingqing:
I restarted build multiple times and it consistently failed on ubuntu build.

korenhe:
ok, I am not quite sure about this patch, someone say this antlr4 is an experiment feature and may have bugs, I can use this patch locally and paste it to a new issue so other Mac users can refer to it. without this patch， mac will generate java result files occasionally.

________________________________

PR #167, opened 2020-07-27
Modify redistribution execution

arzuschen:
Make redistribution to distribute by the target expressions.

Previously, redistribution automatically redistribute by the first expression from child output (r[0]). 
Now the distributeby expression list is extracted from parent join logic node and record in logic redistribute node. This information is later passed to physic node execution. Redistribution is for list of distribution.

### DIFF
--- a/test/UnitTest.cs
+++ b/test/UnitTest.cs
@@ -2090,6 +2090,18 @@ namespace qpmodel.unittest
             Assert.AreEqual(1, TU.CountStr(phyplan, "Gather"));
             Assert.AreEqual(6, TU.CountStr(phyplan, "Redistribute"));
             Assert.AreEqual(1, TU.CountStr(phyplan, "70 threads"));
+
+            // ensure redistribution can shuffle by expression
+            sql = "select a2, b2 from ad, bd where a2*2+a1=b2 order by a2;";
+            TU.ExecuteSQL(sql, "1,2", out phyplan);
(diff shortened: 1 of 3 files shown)


zhouqingqing:
do you want to add any unittest?

test/UnitTest.cs
```diff
@@ -2090,6 +2090,18 @@ public void Redistribute()
...
+            sql = "select a2, b2 from ad, bd where a2*2+a1=b2 order by a2;";
+            TU.ExecuteSQL(sql, "1,2", out phyplan);
+            Assert.AreEqual(1, TU.CountStr(phyplan, "Gather"));
+            Assert.AreEqual(2, TU.CountStr(phyplan, "Redistribute"));
+
+            // no output if by previous r[0] method for redistribution
+            sql = "select d2, a1 from ad, dd where d3=a1 order by d2;";
+            TU.ExecuteSQL(sql, "1,2", out phyplan);
+            Assert.AreEqual(1, TU.CountStr(phyplan, "Gather"));
+            Assert.AreEqual(2, TU.CountStr(phyplan, "Redistribute"));
```
zhouqingqing, line 2104: next step shall remove one redistribute?

arzuschen: Sure, I will remove that. After memo optimization is complete, the necessary redistribution will be removed naturally.

PR #169, opened 2020-07-29
Remove unnecessary redistribution node

arzuschen:
Previously join force redistribution on top of table scan for distributed table, but it is not necessary when the key match the distribution.
Now the match is checked, only when the data distribution does not agree with the join keys, a redistribution node is added. 

Also, data insertion is modified. Previously, even for distributed table all the rows are added onto the first partition (0), but with distributed table scan all have redistribution node on top, it is still working. Now redistribution may not be present, the original insertion must follow the distributed column.

Additional unittest is added to verify the change.

### DIFF
--- a/qpmodel/DataType.cs
+++ b/qpmodel/DataType.cs
@@ -260,6 +260,7 @@ namespace qpmodel.expr
             tableSample_ = tableSample;
         }
 
+        public bool IsDopMatch(QueryOption option) => Table().distributions_.Count == option.optimize_.query_dop_;
         public bool IsDistributed() => Table().distributedBy_ != null;
         public bool IsDistributionMatch(List<Expr> keys, QueryOption option)
         {
@@ -268,7 +269,7 @@ namespace qpmodel.expr
                 return true;
             else
             {
-                Debug.Assert(Table().distributions_.Count == option.optimize_.query_dop_);
+                Debug.Assert(IsDopMatch(option));
                 // table distribution is by one column, check match
                 if (keys.Count == 1 && keys[0] is ColExpr key)
                     if (key.colName_ == Table().distributedBy_.name_)
--- a/qpmodel/DataType.cs
+++ b/qpmodel/DataType.cs
@@ -261,6 +261,20 @@ namespace qpmodel.expr
         }
 
         public bool IsDistributed() => Table().distributedBy_ != null;
+        public bool IsDistributionMatch(List<Expr> keys)
+        {
+            // check if the list of expressions match
+            // the distribution of this table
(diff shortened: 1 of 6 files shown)


arzuschen:
updated to take in consideration for different query dop.
Since the dop for data insertion and dop for query may not be the same, need to force redistribution if the partition number is not the same.

qpmodel/DataType.cs
```diff
@@ -261,6 +261,23 @@ public BaseTableRef(string name, string alias = null, SelectStmt.TableSample tab
         }
 
         public bool IsDistributed() => Table().distributedBy_ != null;
+        public bool IsDistributionMatch(List<Expr> keys, QueryOption option)
+        {
+            // check if the list of expressions match
+            // the distribution of this table
+            if (!IsDistributed())
```
zhouqingqing, line 268: please clarify this comment: only distributed deployment shall check it

qpmodel/DataType.cs
```diff
@@ -261,6 +261,23 @@ public BaseTableRef(string name, string alias = null, SelectStmt.TableSample tab
...
+                return true;
+            else
+            {
+                // check if number of partitions agree
+                if (option.optimize_.query_dop_ == Table().distributions_.Count)
```
zhouqingqing, line 273: let's assert dop = distribution count

qpmodel/LogicNode.cs
```diff
@@ -585,27 +586,31 @@ void createOneKeyList(BinExpr fb)
                 Debug.Assert(leftKeys_.Count == rightKeys_.Count);
             }
 
-            // reset the old values
-            if (!canReUse)
+            // make sure multithreading does not mess up shared logic node
```
zhouqingqing, line 589: can you clarify this?

qpmodel/LogicNode.cs
```diff
@@ -585,27 +586,31 @@ void createOneKeyList(BinExpr fb)
...
+            lock (this)
```
zhouqingqing, line 590: adjust order to avoid this?

qpmodel/PhysicalNode.cs
```diff
@@ -1431,12 +1431,20 @@ public class PhysicInsert : PhysicNode
 
         public override string Exec(Func<Row, string> callback)
         {
+            var table = (logic_ as LogicInsert).targetref_.Table();
+            var dop = context_.option_.optimize_.query_dop_;
+            if (table.distributedBy_ != null)
```
arzuschen, line 1436: put assertion in create table ref

Issue #125, opened 2020-06-20
change Exec()/Close() codegen to use ExecContext.code_ 


zhouqingqing:
See f744db8b80cca2ed8a2ff0a7ad8d667779408c43

zhouqingqing:
fixed

PR #171, opened 2020-08-04
change Exec()/Close() codegen to use ExecContext.code_ #125

zhouruiapple:
fix issues from code factor.
generate code to @context.code_ directlly.
change the paramter of Exec from Func to Action.
all Exec() of query plans do not return generated code.

### DIFF
--- a/qpmodel/ExprSubquery.cs
+++ b/qpmodel/ExprSubquery.cs
@@ -173,7 +173,6 @@ namespace qpmodel.expr
             {
                 // exists check can immediately return after receiving a row
                 r = l;
-                return null;
             });
 
             bool exists = r != null;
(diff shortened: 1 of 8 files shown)


zhouqingqing:
1. please fix codefactor 4 issues;
2. please investigate why coverage drops 0.4%.

zhouruiapple:
The code coverage rate decreases due to the following two reasons:

1. Brackets are added to the conditional branches of a single line. For example [https://coveralls.io/builds/32545682/source?filename=qpmodel%2FPhysicalNode.cs#L356](url).
2. Move the code of codegen_if to caller to generate code to @context.code_. For example [https://coveralls.io/builds/32545682/source?filename=qpmodel%2FPhysicalNode.cs#L612](url).

we need to add test case to cover above code. @zhouqingqing

zhouqingqing:
> The code coverage rate decreases due to the following two reasons:

That's ok - as long as we understand the reasons.

Issue #45, opened 2020-01-20
SQL requires case insensitive names


zhouqingqing:
we handled some but not all

PR #175, opened 2020-08-05
make SQL identifiers case insensitive (issue #45)

pkommoju:


zhouqingqing:
@pkommoju  fix the codefactor issues please

qpmodel/Catalog.cs
```diff
@@ -51,7 +51,11 @@ public class ColumnDef
         readonly public ColumnType type_;
         public int ordinal_;
 
-        public ColumnDef(string name, ColumnType type, int ord) { name_ = name; type_ = type; ordinal_ = ord; }
+        public ColumnDef(string name, ColumnType type, int ord)
+        {
+            name_ = name.StartsWith('"') ? name : name.ToLower();
```
zhouqingqing, line 56: can we add one utility function 
string normalizeName(string name) => name.StartsWith('"') ? name : name.ToLower();

to do the conversion?

pkommoju: Sure. Let me do it right away. Will adding it in Utils.cs work?

pkommoju: Looks like all source files use qpmodel.expr and expr.cs is used by all except PrestoStats.cs. So that should work.

zhouqingqing:
can you fix the code factor issues?

PR #176, opened 2020-08-07
modify jobench and add unittest

arzuschen:
Remove unnecessary code in jobench unittest and add jobench stats import for program.cs.
Add codegen unittest to cover left NLJoin, antisemi NLJoin and null groupby_ for aggregation.

### DIFF
--- a/test/UnitTest.cs
+++ b/test/UnitTest.cs
@@ -225,6 +225,12 @@ namespace qpmodel.unittest
             sql = "select a2*2, count(a1) from a, b, c where a1=b1 and a2=c2 group by a2 limit 2;";
             TU.ExecuteSQL(sql, "2,1;4,1", out phyplan, option);
 
+            sql = "select a1, b1 from a left join b on a1>b1;";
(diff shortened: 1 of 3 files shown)


test/UnitTest.cs
```diff
@@ -297,10 +302,6 @@ public void TestJobench()
             string write_dir_fn = $"../../../../test/regress/output/jobench";
             string expect_dir_fn = $"../../../../test/regress/expect/jobench";
 
-            // make sure all queries can generate phase one opt plan
-            QueryOption option = new QueryOption();
-            option.optimize_.TurnOnAllOptimizations();
```
zhouqingqing, line 302: Please keep this line as it turns on all optimizations for jobench run.

arzuschen: I think this option is not used, the option has optimize turned on in psql/Program.cs line 31

zhouqingqing: You are right. In a separate patch, can you please remove psql project and move its content to unittest? We used to have psql designed as a SQL client but we don't think we will need it anymore.

arzuschen: got it

Issue #183, opened 2020-08-11
type coerce issues


zhouqingqing:
We have some type coerce code in place but needs strengthen it.

Tasks:
1. add more test cases covering type coerce;
 - operators like +/-/like
 - functions like year/Aggs
 - built in language like UNION, CASE
2. bug fixes based with above test cases.

Refs:
[SQL Server data types conversion rules](https://docs.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql?view=sql-server-ver15) 
[PostgreSQL data types conversion algorithms](https://www.postgresql.org/docs/current/typeconv.html)

Known issues:
- sql = "select year(date('2011-01-01'));";

zhouqingqing:
The known issue `select year(date('2011-01-01'))` is now fixed in PR #319:

1. `year()` return type changed from `DateTimeType` to `IntType` (commit ac69d59)
2. `year()` exec now handles both DateTime and string inputs
3. `abs()` now preserves integer type instead of converting to decimal (commit 03e7749)

```
> select year(date('2011-01-01'))
2011
```

zhouqingqing:
The remaining type coerce improvements (beyond the year(date()) fix) are enhancement work rather than specific bugs. Keeping open for tracking.

PR #184, opened 2020-08-11
introduce type precedence table to replace hard wired type conversion logic

zhouqingqing:
see details in #183

### DIFF
--- a/qpmodel/ExprFunc.cs
+++ b/qpmodel/ExprFunc.cs
@@ -744,7 +744,7 @@ namespace qpmodel.expr
             base.Bind(context);
             type_ = then_()[0].type_;
             if (else_() != null)
-                Debug.Assert(type_.Compatible(else_().type_));
+                Debug.Assert(TypeBase.Compatible(type_, else_().type_));
         }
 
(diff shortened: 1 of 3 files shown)


qpmodel/ExprFunc.cs
```diff
@@ -744,7 +744,7 @@ public override void Bind(BindContext context)
             base.Bind(context);
             type_ = then_()[0].type_;
             if (else_() != null)
-                Debug.Assert(type_.Compatible(else_().type_));
+                Debug.Assert(TypeBase.Compatible(type_, else_().type_));
         }
```
pkommoju, line 748: This assert will always be TRUE (never fails) because TypeBase.Compatible() is returning true unconditionally. Do you intend to add actual compatibility checks in TypeBase.Compatible() later, or the actual compatibility is determined elsewhere?

zhouqingqing: It is true because we haven't implemented this functionality. This functionality shall be done as part of #183

pkommoju: Got it.

PR #187, opened 2020-08-14
Add TPCH distributed unittest

arzuschen:
Similar to TPCH0001, all the tables have their distributed/replicated counterpart created. Generally, small tables are replicated and others are distributed on the key. The tables are named tablename_r and tablename_d respectively.
q02, q04, q07, q08, q09, q13, q14, q15, q17, q18, q20, q21, q22 cannot be run at this point.


tpch/distributed_queries/q01.sql
```diff
@@ -0,0 +1,23 @@
+-- using default substitutions
```
zhouqingqing, line 1: we don't have to copy queries again - they are supposed to be the same as non-distributed queries

arzuschen: changed back to the original table names

PR #191, opened 2020-08-17
fix race condition of distributed execution

zhouqingqing:
In the real execution, we  will have to serialize all information we want the receiver side to deserialize, including physic node an logic node and could be some more. So to emulate this behavior correctly, we shall either having one copy for each participanting thread or make sure their shared copy not change. We do the second way by moving CreateKeyList() to the last step in logic join creation, instead of every physic join doing that separately.

### DIFF
--- a/qpmodel/LogicNode.cs
+++ b/qpmodel/LogicNode.cs
@@ -366,7 +366,9 @@ namespace qpmodel.logic
         // 2. compute children's output by requesting reqOutput from them
         // 3. find mapping from children's output
         //
-        // this parent class implments a default one for using first child output without any change
+        // this parent class implments a default one for using first child output without any change. You can
+        // also piggy back in this method to "finalize" your logic node.
+        //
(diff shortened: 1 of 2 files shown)


zhouqingqing:
2020-08-17T19:14:52.4434000Z Debug Trace:
2020-08-17T19:14:52.4434251Z ../../../../test/regress/output/tpch0001_d/q16.txt;
2020-08-17T19:14:52.4434397Z 
2020-08-17T19:14:52.4434569Z 
2020-08-17T19:15:33.0668374Z 
2020-08-17T19:15:33.0682603Z Test Run Failed.
2020-08-17T19:15:33.0682859Z Total tests: 47
2020-08-17T19:15:33.0683217Z      Passed: 46
2020-08-17T19:15:33.0683450Z      Failed: 1

q16 failure is a known issue due to subquery execution and not related to this change.

PR #194, opened 2020-08-18
Add distribution to physic property

arzuschen:
Distribution attribute is added into physic property. Since in the new framework, all the requirements are regarded as top-to-down requirements, there are 4 different types of distribution requirements:
1) Singleton: not distributed execution
2) Distributed (on Expr list): need to be hash distributed according to certain list of expressions
3) Replicated: data stream is replicated across the threads
4) Any: can be supplied by any distribution (i.e., anything will work)

The default base requirement is singleton and any order posed on top. For query concerning distributed tables, logic tree comes with remote exchange nodes like redistribution and gather. For memo optimization, these nodes are removed from the logic tree and distribution transformation is enabled. Previously, all distributed queries are blocked from using memo, now with an additional option parameter memo_use_remoteexchange_, they can be optimized by memo, and no change in the calculation for non-distributed queries.
There are two special cases for distribution. Replicated distribution is not considered in Any since it uses a different gather enforcement node. For a hash join, a replicated requirement results in replicated requirement on both children. There is also a round robin for table data distribution, it is considered as distributed with empty expression list. It will need to be redistributed if any particular distributed is required but can satisfy any.

For example, this query: `select a1, count(b1) from ad, b where a1 > b2 group by a1 order by a1`
In this case, ad is distributed on a1. The base required property is <a1-asc, singleton> and the root group is aggregate.
<a1-asc, singleton> is then transformed in to <none, singleton>, and that is further transformed into <none, any> and <none, replicated.
Assume hashagg and streamagg is only compatible with singleton, so <none, any> and <none, replicated> cannot be supported. For this group, it may be that <a1-asc, singleton> is directly supported by streamagg, and same property requirement posed on the child group, and <none, singleton> may be supported by hashagg, the requirement on child group is then <none, singleton>.
Consider join group after that similarly, the requirements are transformed to get <none, any> and <none, replicated>. For the case of hashjoin, <none, any> required on top will result in 4 different set of child requirements:
{<none, singleton>, <none, dist on b2>} {<none, dist on a1>, <none, singleton>} {<none, dist on a1>, <none, dist on b2>} {<none, replicated>, <none, dist on b2>}
For {<none, singleton>, <none, singleton>}, the result is singleton, that situation should be considered in the <none, singleton> requirement. Also, the join order is not demonstrated since it is included in the logic join transformation rules.
If there are more than one set of child requirements, the inclusive cost for each set is calculation and the least-cost set is select. In this case, it may be that {<none, dist on a1>, <none, singleton>} is the best set since no enforcement will be needed for the table scan.

### DIFF
--- a/test/regress/expect/tpch0001_d/q03.txt
+++ b/test/regress/expect/tpch0001_d/q03.txt
@@ -21,35 +21,35 @@ order by
 	revenue desc,
 	o_orderdate
 limit 10
-Total cost: 23009.8, memory=267728
-PhysicLimit (10) (inccost=23009.8, cost=10, rows=10) (actual rows=8)
+Total cost: 28022.9, memory=62600
+PhysicLimit (10) (inccost=28022.9, cost=10, rows=10) (actual rows=8)
     Output: l_orderkey[0],{sum(l_extendedprice*1-l_discount)}[1],o_orderdate[2],o_shippriority[3]
-    -> PhysicOrder  (inccost=22999.8, cost=2851.9, rows=458, memory=10992) (actual rows=8)
+    -> PhysicOrder  (inccost=28012.9, cost=2851.9, rows=458, memory=10992) (actual rows=8)
         Output: l_orderkey[0],{sum(l_extendedprice*1-l_discount)}[1],o_orderdate[2],o_shippriority[3]
         Order by: {sum(l_extendedprice*1-l_discount)}[1], o_orderdate[2]
-        -> PhysicHashAgg  (inccost=20147.89, cost=1374, rows=458, memory=21984) (actual rows=8)
+        -> PhysicHashAgg  (inccost=25161, cost=1374, rows=458, memory=21984) (actual rows=8)
             Output: {l_orderkey}[0],{sum(l_extendedprice*1-l_discount)}[3],{o_orderdate}[1],{o_shippriority}[2]
             Aggregates: sum(l_extendedprice[4]*1-l_discount[7])
             Group by: l_orderkey[0], o_orderdate[1], o_shippriority[2]
-            -> PhysicGather 1 : 10 (inccost=18773.89, cost=45.8, rows=458) (actual rows=14)
+            -> PhysicGather 1 : 10 (inccost=23787, cost=4580, rows=458) (actual rows=14)
                 Output: l_orderkey[2],o_orderdate[3],o_shippriority[4],{l_extendedprice*1-l_discount}[5],l_extendedprice[6],{1-l_discount}[7],{1}[0],l_discount[8]
-                -> PhysicHashJoin  (inccost=18728.09, cost=2097, rows=458, memory=464) (actual rows=0)
+                -> PhysicHashJoin  (inccost=19207, cost=2097, rows=458, memory=464) (actual rows=0)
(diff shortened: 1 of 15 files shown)


test/regress/expect/tpch0001_d/q10.txt
```diff
@@ -30,45 +30,61 @@ group by
...
-                        -> PhysicScanTable customer (inccost=150, cost=150, rows=150) (actual rows=0)
-                            Output: c_custkey[0],c_name[1],c_acctbal[5],c_address[2],c_phone[4],c_comment[7],1,c_nationkey[3]
-                    -> PhysicRedistribute  (inccost=13172, cost=150, rows=1500) (actual rows=0)
-                        Output: n_name[0],{l_extendedprice*1-l_discount}[1],l_extendedprice[2],{1-l_discount}[3],l_discount[4],o_custkey[5],n_nationkey[6]
-                        -> PhysicNLJoin  (inccost=13022, cost=2450, rows=1500) (actual rows=0)
-                            Output: n_name[0],{l_extendedprice*1-l_discount}[2],l_extendedprice[3],{1-l_discount}[4],l_discount[5],o_custkey[6],n_nationkey[1]
-                            -> PhysicScanTable nation (inccost=25, cost=25, rows=25) (actual rows=0)
-                                Output: n_name[1],n_nationkey[0]
-                            -> PhysicRedistribute  (inccost=10547, cost=6, rows=60) (actual rows=0)
+            -> PhysicGather 1 : 10 (inccost=11111, cost=900, rows=90) (actual rows=140)
```
zhouqingqing, line 43: why we have a Gather node here? It is essentially serialize the plan.

arzuschen: currently I made hashagg only able to deal with singleton, therefore gather will be enforced below hashagg.
I tried enabling distribution for hashagg and gather can be place on top of hashagg, but it will result in inconsistent results for average expression.

test/regress/expect/tpch0001_d/q03.txt
```diff
@@ -21,35 +21,35 @@ order by
...
             Aggregates: sum(l_extendedprice[4]*1-l_discount[7])
             Group by: l_orderkey[0], o_orderdate[1], o_shippriority[2]
-            -> PhysicGather 1 : 10 (inccost=18773.89, cost=45.8, rows=458) (actual rows=14)
+            -> PhysicGather 1 : 10 (inccost=23787, cost=4580, rows=458) (actual rows=14)
                 Output: l_orderkey[2],o_orderdate[3],o_shippriority[4],{l_extendedprice*1-l_discount}[5],l_extendedprice[6],{1-l_discount}[7],{1}[0],l_discount[8]
-                -> PhysicHashJoin  (inccost=18728.09, cost=2097, rows=458, memory=464) (actual rows=0)
+                -> PhysicHashJoin  (inccost=19207, cost=2097, rows=458, memory=464) (actual rows=0)
                     Output: l_orderkey[2],o_orderdate[3],o_shippriority[4],{l_extendedprice*1-l_discount}[5],l_extendedprice[6],{1-l_discount}[7],{1}[0],l_discount[8]
                     Filter: c_custkey[1]=o_custkey[9]
                     -> PhysicScanTable customer (inccost=150, cost=150, rows=29) (actual rows=0)
```
zhouqingqing, line 39: here we choose to redistribute join 1581 rows instead of broadcast/redistribute customer as build side?

arzuschen: I didn't have redistribution enforcement yet, I will add that

Issue #173, opened 2020-08-04
[bug] race condition in emulated distributed execution


zhouqingqing:
Test case Redistribute(), which does not reliably repro but you can make it easy to repro with a endless loop:

```
       public void Redistribute()
        {
            for (; ; )
            {
                // needs order by to force result order
                var phyplan = "";
```

when it is fixed, change for(;;) to 3 times.

zhouqingqing:
@zhouruiapple updates?

zhouruiapple:
@zhouqingqing In progressing right now

zhouruiapple:
@zhouqingqing Can you explain the bug with more information? I don't find the race condition in the code of Redistribute().

zhouqingqing:
Have you tried putting the loop around the unittest? -------- Original message --------From: zhourui <notifications@github.com>Date: Mon, Aug 10, 2020, 19:31To: zhouqingqing/qpmodel <qpmodel@noreply.github.com>Cc: Qingqing Zhou <zhouqq.github@gmail.com>, Mention <mention@noreply.github.com>Subject: Re: [zhouqingqing/qpmodel] [bug] race condition in emulated distributed execution (#173)
@zhouqingqing Can you explain the bug with more information? I don't find the race condition in the code of Redistribute().

zhouqingqing:
another issue is profiling nodes actual rows = 0.

Check out here:
https://github.com/zhouqingqing/qpmodel/pull/187/files

Look at tpch query plans with actual stats. You can see many with "actual rows = 0" under gather. I think what's happened here is that each individual threads have do their profiling correctly but there is no process to "summarize" profilings into the main thread's profiling nodes, so we see all zeros.

zhouqingqing:
another issue by testing TestTpchDistributed()

Put a for(;;) loop around line 413, you shall repro an issue with the cached subquery

(actual rows, loops = 2)

where loops = 2 is extra garbage. Most likely Q16.

zhouruiapple:
@zhouqingqing Got it.

zhouruiapple:
The @totalfreq_ of @MCVList of ColumnStat cause EstSelectivity which return different value when unit test issued. Is it changed in different threads? @zhouqingqing

zhouqingqing:
check out how ANALYZE is implemented - there is a GATHER on top of select statement, so analyze itself is not multi-threaded.

zhouruiapple:
[...]

PR #198, opened 2020-08-19
support distributed query profiling

zhouruiapple:
…thread #173

assign aggregation profiling to cloned profiling.

### DIFF
--- a/qpmodel/ExprSubquery.cs
+++ b/qpmodel/ExprSubquery.cs
@@ -279,7 +279,7 @@ namespace qpmodel.expr
 
             cachedVal_ = set;
             cachedValSet_ = true;
-            return set.Contains(expr); ;
+            return set.Contains(expr);
         }
     }
 
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -1695,6 +1695,24 @@ namespace qpmodel.physic
             Debug.Assert(profile_ is null);
         }
 
+        // Use atomic variable instead of mutex to ensure
+        // that modifications to @nrows_ and @nloops_ are atomic.
+        // This is the simplest lock-free idiom.
+        //
+        // The advantage is that there is no overhead
+        // for thread switching even in distributed query.
+        //
+        // The disadvantage is that non-distributed query
+        // also has atomic variable synchronization overhead.
+        // Since atomic variable does not cause thread switching,
+        // overhead is almost negligible.
+        //
+        // We can also distinguish the current query mode by @context_.
+        // If @context_ is @DistributedContext, increase @nrows_ and @nloops_.
+        // Otherwise increase @nrows_ and @nloops_ atomaticlly.
+        // In this situation, duplicate codes are generated.
+        //
+        // Let us use atomic variable for both case now.
         public override void Exec(Action<Row> callback)
         {
             ExecContext context = context_;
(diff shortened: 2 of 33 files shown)


zhouruiapple:
There is still a issue of race condiation, which will be solved later.

qpmodel/ExprSubquery.cs
```diff
@@ -261,25 +261,28 @@ public override void Bind(BindContext context)
 
         public override Value ExecNonDistributed(ExecContext context, Row input)
         {
-            Debug.Assert(type_ != null);
-            Value expr = expr_().Exec(context, input);
+            lock(this)
```
zhouqingqing, line 264: why this change? Do we already handled this in Exec()?

qpmodel/stmt.cs
```diff
@@ -131,6 +131,13 @@ public virtual List<Row> Exec()
             {
                 Console.WriteLine(physicPlan_.Explain(queryOpt_.explain_));
             }
+
+            var dcontext = context as DistributedContext;
+            if (dcontext != null)
+            {
+                dcontext.machines_.WaitThreads();
```
zhouqingqing, line 138: why we need to waitThreads() here?

qpmodel/ExprSubquery.cs
zhouruiapple: The mainline code also fixes the bug. Keep the bug fix code in this commit the when merging code.
I will remove above code.

test/regress/expect/tpch0001_d/q01.txt
```diff
@@ -29,7 +29,7 @@ PhysicOrder  (inccost=71071.35, cost=11.35, rows=6, memory=372) (actual rows=4)
         Group by: l_returnflag[0], l_linestatus[1]
         -> PhysicGather 1 : 10 (inccost=65135, cost=59130, rows=5913) (actual rows=5914)
             Output: l_returnflag[8],l_linestatus[9],l_quantity[4],l_extendedprice[5],l_extendedprice[5]*1-l_discount[6],1-l_discount[6],1,l_discount[6],l_extendedprice[5]*1-l_discount[6]*1+l_tax[7],1+l_tax[7],l_tax[7]
-            -> PhysicScanTable lineitem (inccost=6005, cost=6005, rows=5913) (actual rows=0)
+            -> PhysicScanTable lineitem (inccost=6005, cost=6005, rows=5913) (actual rows=591, loops=10)
```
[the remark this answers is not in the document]
zhouruiapple: Got it. Can you create a issue for this change?

qpmodel/stmt.cs
zhouruiapple: If the query plan(in LISP) as (PhysicGather(...(PhysicGather))), we must wait all threads to guarantee loop count of second PhysicGather is fixed value.

qpmodel/PhysicalNode.cs
```diff
@@ -1697,17 +1707,33 @@ public override void Exec(Action<Row> callback)
...
+                    {
+                        aggProfiling_.nrows_++;
+                    }
                     callback(l);
                 }
             });
         }
 
+        // assign aggregation profiling to cloned profiling.
+        public override PhysicNode Clone()
```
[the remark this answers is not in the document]
zhouruiapple: Query plan under Gather with actual "actual rows = 0" because of all Profilings under Gather are cloned. So we should keep reference of original Profiling in all cloned profilings.

All expected results are updated.

qpmodel/Statis.cs
```diff
@@ -346,7 +346,8 @@ public void ComputeStats(int index, List<Row> samples)
                 foreach (var g in groups)
                     sortgroup.Add(g.Key, g.Count());
 
-                var sorted = from pair in sortgroup orderby pair.Value descending select pair;
+                // use top 100 values to calculate frequency, ensure that pairs are sorted in a fixed order.
+                var sorted = from pair in sortgroup orderby pair.Value descending, pair.Key descending select pair;
```
zhouruiapple, line 350: we must keep the order of @sorted. Sort by @pair.Value, and then sort by @pair.Key. Otherwise top 100 of @sorted cant not be guaranteed.

PR #199, opened 2020-08-19
Enable replicated enforcement and implement broadcast execution

arzuschen:
This patch consists of two main parts:
1) Enable replicated enforcement
Now replicated distribution can be enforced from distribution on any list expression through broadcast node. This make it possible for hashjoining two distributed tables by broadcasting the smaller table for the build end. Therefore the total cost may decrease since a join previously requiring two redistribution can now be replaced by only one broadcast.
2) Implement broadcast execution
Broadcast physic node execution is implemented using shared code from redistribution. The class of Physic RemoteExchange is modified to adopt to the shared portion for broadcast and redistribution. The estimated cost for broadcast is set to the same as redistribution, which is 2.0x the cardinality of child node.

After the above is implemented, the plans for redistribution unittest changes since using broadcasting is likely to produce lower cost plans. Therefore, a parameter in option.optimize is introduced to turn on/off the replicated distribution enforcement. This way the plans for redistribution test does not need to change.
Distributed TPCH result plans are changed accordingly.

### DIFF
--- a/qpmodel/PhysicalNode.cs
+++ b/qpmodel/PhysicalNode.cs
@@ -198,16 +198,24 @@ namespace qpmodel.physic
 
         // interface for physical property
         //
-        // what property current node requires to implment it
-        //      for example, StreamAgg requires Sort on grouping keys
-        public virtual PhysicProperty RequiredProperty() => new PhysicProperty();
-        // what property current node can supply
-        //      for example, StreamAgg can supply Sort on grouping keys
-        public virtual PhysicProperty SuppiedProperty() => new PhysicProperty();
-        // requirements of each children node if propogate the property
-        //      for example, NLJ sort order requires outer node the same order
-        public virtual List<PhysicProperty> PropagatedProperty(PhysicProperty property)
-            => new List<PhysicProperty>(new PhysicProperty[children_.Count]);
+        // determine if a required property can be satisfied through physic node
+        // if it can be satisfied, then property requirement on children is returned
+        public virtual bool IsPropertySatisfied(PhysicProperty required, out List<PhysicProperty> childprops)
+        {
+            // when there is no property requirement, it is always satisfied
+            if (required.Equals(PhysicProperty.nullprop))
+            {
+                childprops = new List<PhysicProperty>();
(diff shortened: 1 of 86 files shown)


qpmodel/PhysicalNode.cs
```diff
@@ -1904,8 +1907,47 @@ public PhysicNode EmulateSerialization(PhysicNode node)
             var newnode = node.Clone();
             return newnode;
         }
-        public virtual string OpenConsumer(ExecContext context) => null;
-        public virtual string OpenProducer(ExecContext context) => null;
+        public virtual string OpenConsumer(ExecContext econtext)
```
zhouqingqing, line 1910: I would suggest you keep OpenConsumer() as it is but provides an helper function say createProducersAndChannels() for Redistribute/Broadcast to reuse.

Same as other virtual functions.

arzuschen: got it

test/UnitTest.cs
```diff
@@ -2256,73 +2256,75 @@ public void Gather()
         [TestMethod]
         public void Redistribute()
         {
+            var option = new QueryOption();
+            option.optimize_.enable_broadcast_ = false;
```
zhouqingqing, line 2260: how about adding a loop on turn on/off enable_broadcast_ and assert either way the results are the same. If the plan is different, assert the difference. See similar usage in tpch tests.

test/regress/expect/tpch0001_d/q16.txt
```diff
@@ -28,33 +28,33 @@ order by
 	p_brand,
 	p_type,
 	p_size
-Total cost: 6300.38, memory=22274
-PhysicOrder  (inccost=6300.38, cost=754.38, rows=148, memory=6364) (actual rows=0)
+Total cost: 4774.38, memory=22274
```
zhouqingqing, line 31: good - at least we see with broadcast in, all new cost are down!

test/UnitTest.cs
arzuschen: sure, can I create a broadcast test method to do that?

PR #203, opened 2020-08-20
Gatherexplainfix

zhouqingqing:


qpmodel/ExprSubquery.cs
```diff
@@ -36,6 +36,7 @@
 using qpmodel.logic;
 using qpmodel.physic;
 using qpmodel.utils;
+using IronPython.Compiler.Ast;
```
arzuschen, line 39: Is this necessary? I can't find where it is used

zhouqingqing: interesting, not sure when it get sneaked in - thanks for spotting it

PR #206, opened 2020-08-25
expression normalization

pkommoju:
Semantic Normalization and Transformation changes to produce a canonical
form of the statement and eliminate unnecessary operation.

The module is called Normalizer and it does the following transformations:
1) Constant move. Bring all possible constants together so that later
   transformations can simplify or even remove some of the constants.
   Ex. CONST + expr => expr + CONST

2) Constant folding. Replace expressions involving constants with the value of that part of the expression. This includes some aggregates and functions too.
   Ex. CONST op CONST  => the value after performing the operation.

3) The Arithmetic Simplification. Slightly specialized for of constant folding.
     Ex. expr + 0, expr - 0, expr * 1, expr / 1 are replaced by expr.

4) Comparison Simplification. Like constant folding but or comparison. It covers more than
    expr > const, expr = expr being replaced by TRUE or FALSE. Some of these transformations
    are not done yet.

5)  Logical operation simplification, including logical operations involving expression know to be
     TRUE or FALSE. The biggest thing not done involves NOT. NOT (col > 10) should be replaced
     by col <= 10.

6)  IN list simplification, DISTINCT elimination, redundant CAST elimination, CASE simplification
     are not done.

All classes derived from Expr may override Normalize() method to customize normalization.
Normalizer runs after Binder. There is a new method Expr.BindAndNormalize() which calls bind, and then normalizer.

Not all transformations are attempted, some should not be done at all ( const - col is not the same
as col - const, and transforming division is avoided because of he chance of loss of precision).

All classes affected by Normalizer are redefined as partial and their Normalize() methods put
into Normalizer.cs.

TODO
1) CASE: case handling is incorrect, so that needs to be fixed first.
2) logical simplification is not thorough. For instance, where c1 <> c1 should be reduced to WHERE FLSE,  and, where (a1 = a2) AND (a3 = a3) should be eliminated.
3) CAST elimination.
4) DISTINCT elimination: min(distinct 10) is simply 10 but some optimizer changes may also be required for this to happen.


qpmodel/Expr.cs
```diff
@@ -1225,6 +1238,44 @@ public static ConstExpr MakeConst(Value val, ColumnType type, string outputName
         }
         public static ConstExpr MakeConstBool(bool istrue) =>
             istrue ? MakeConst("true", new BoolType()) : MakeConst("false", new BoolType());
+
+        public bool IsNull() => val_ is null;
+
+        public bool IsZero()
+        {
+
```
zhouqingqing, line 1246: please remove all extra empty lines

qpmodel/Program.cs
```diff
@@ -138,6 +176,8 @@ static void Main(string[] args)
                 sql = args[0];
             }
 
+#pragma warning disable CS0162 // Unreachable code detected
+            // Prasad Kommoju: The warnings are annoying, so using a pragma to suppress them.
```
zhouqingqing, line 180: please remove names here

pkommoju: Got it.

qpmodel/Expr.cs
pkommoju: Yes, I will do that in the next push.

qpmodel/stmt.cs
```diff
@@ -626,6 +626,11 @@ LogicNode FilterPushDown(LogicNode plan, bool pushJoinFilter)
                     // it there - shall we try hard to stop query early? Nope, it is no deserved
                     // to poke around for this corner case.
                     //
+                    // One of the expression normalization rules replaces terms like
+                    // COL > NULL, COL < NULL COL <> NULL with a FALSE term and
+                    // if all of the terms where reduced to WHERE FALSE
+                    // the above policy and assert are no longer valid.
```
zhouqingqing, line 632: "assert are no longer valid" which assertion - do you mean " Debug.Assert(trueOrFalse)"? If that's the case, you don't mention that in your new code comment.

L649: so trueOrFalse can be false, what if it is false?

qpmodel/stmt.cs
```diff
@@ -671,15 +676,15 @@ public bool SubqueryIsWithMainQuery(NamedQuery subquery)
         {
             // FromQuery or decorrelated subqueries are merged with main plan
             var r = (fromQueries_.ContainsKey(subquery) ||
-                decorrelatedSubs_.Contains(subquery));
```
zhouqingqing, line 674: are these changes from auto-format? please keep the indent space - the indent space meaning the new line  is not a new statement, but a continuation.

qpmodel/stmt.cs
pkommoju: I actually meant to change the original comment to "if it is a constant true filer, remove it. If a false filter, push it down if possible to help the optimizer eliminate the query node (scan etc) if possible. But left the original comment and added an explanation for your review before finalizing the decision to push a false filter or not.

qpmodel/stmt.cs
pkommoju: I will run the auto-format.

pkommoju: I think this line breaking is accidental because it is not part of normalization.

pkommoju: I tried "format selection" but it did nothing. I will join the lines, is that alright?

test/UnitTest.cs
```diff
@@ -1412,13 +1412,12 @@ public void TestConst()
             sql = "select 1+2*3, 1+2+a1 from a where a1+2+(1*5+1)>2*3 and 1+2=2+1;";
             TU.ExecuteSQL(sql, "7,3;7,4;7,5", out phyplan);
             Assert.AreEqual(0, TU.CountStr(phyplan, "True"));
-            Assert.AreEqual(1, TU.CountStr(phyplan, "7,3+a.a1[0]"));
-            Assert.AreEqual(1, TU.CountStr(phyplan, "a.a1[0]+2+6>6")); // FIXME
+            Assert.AreEqual(1, TU.CountStr(phyplan, "Output: 7,(a.a1[0]+3)"));
+            Assert.AreEqual(1, TU.CountStr(phyplan, "Filter: a.a1[0]>-2"));
```
zhouqingqing, line 1416: say a.a1+b.b1=2: if we can arrange it to a.a1=2-b.b1 then a hash join is applicable otherwise, only NLJ is possible. So we may need a different than "put const on right side" rule to rewrite for join condition?

test/UnitTest.cs
```diff
@@ -1769,16 +1768,16 @@ public void TestSort()
...
-                            Order by: {4-a.a3/2}[0]
-                            -> PhysicHashAgg   (actual rows=2)
-                                Output: {4-a.a3/2}[0],{4-a.a3/2}[0]*2+1+{min(a.a1)}[1],{avg(a.a4)}[2]+{count(a.a1)}[3],{max(a.a1)}[4]+{sum(a.a1+a.a2)}[5]*2
-                                Aggregates: min(a.a1[1]), avg(a.a4[2]), count(a.a1[1]), max(a.a1[1]), sum(a.a1[1]+a.a2[4])
-                                Group by: {4-a.a3/2}[0]
-                                -> PhysicScanTable a  (actual rows=3)
-                                    Output: 4-a.a3[2]/2,a.a1[0],a.a4[3],a.a1[0]+a.a2[1],a.a2[1],a.a3[2]";
+            result = ExecuteSQL(sql, out string phyplan);           
+            var answer = @"PhysicOrder  (actual rows=2)
+                            Output: {(4-a.a3)/2}[0],{(((4-a.a3)/2*2+1)+min(a.a1))}[1],{(avg(a.a4)+count(a.a1))}[2],{(max(a.a1)+sum((a.a1+a.a2))*2)}[3]
```
zhouqingqing, line 1773: ((( -- too many but I guess we have to stay with it for now better than wrong impression of bad maths.

qpmodel/Plan.cs
```diff
@@ -663,7 +663,7 @@ List<Expr> bindSelectionList(BindContext context)
                     x.Bind(context);
                     if (x.HasAggFunc())
                         hasAgg_ = true;
-                    x = x.ConstFolding();
+                    // x = x.ConstFolding();
```
zhouqingqing, line 666: remove it

qpmodel/Plan.cs
```diff
@@ -722,6 +723,8 @@ internal void BindWithContext(BindContext context)
 
             if (orders_ != null)
                 orders_ = bindOrderByOrGroupBy(context, orders_);
+
+            Normalize();
```
zhouqingqing, line 727: is it possible to move Normalize() ridding together with Expr.Bind()?

zhouqingqing:
please take a look at coverage/coveralls - we see -0.6% down. Once source is from Normalizer.cs - please see if you can add more test coverage there.

Issue #207, opened 2020-08-25
[bug] Incorrect handling of case expression


pkommoju:
CASE behavior is not conforming to the standard.

select case when a1 is not null then a1 else a4 end from a;

produces
3
4
5
===

The equivalent construct

select coalesce(a1, a4) from a;

produces correct result
0
1
2

A more illustrative statement:

select case a1 when 0 then 'a' when 1 then 'b' when 2 then 'c' else 'd'  end from a;

produces
a
a
a

In effect only the first WHEN's THEN clause is output unconditionally.

PR #209, opened 2020-08-29
fix incorrect handling of case expression #207

zhouruiapple:
rewrite @VisitCaseExpr of @SQLParser.cs.
add "is not" to nullops of BinExpr.
add unit test for simple case searched case.

### DIFF
--- a/test/UnitTest.cs
+++ b/test/UnitTest.cs
@@ -2373,22 +2373,22 @@ namespace qpmodel.unittest
             // May need to instrument with extra output for CASE expressionsa and add
             // something like "NORMTRAN: Constant case folded"
             // Assert.IsTrue(phyplan.Contains("NORMTRAN: Constant case folded"));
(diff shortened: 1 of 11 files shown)


test/UnitTest.cs
```diff
@@ -2230,6 +2230,65 @@ public void TestCaseSensitivity()
...
+                Assert.AreEqual(3, result.Count);
+                Assert.AreEqual(1, result[0].ColCount());
+                String[] expectRows = {"a", "b", "c"};
+                for (int i = 0; i < result.Count; i++)
+                    Assert.AreEqual(expectRows[i], result[i][0]);
+            }
+
+            {
+                string sql = "select case a1 when a1+a2 then a3 else a4 end from a;";
+                var result = ExecuteSQL(sql);
```
zhouqingqing, line 2250: you can write ExecuteSQL(sql, "3;4;5");

zhouruiapple: Done.

Issue #143, opened 2020-07-01
relaxing assertion newmember memo signature after rule application must equal to existing members


zhouqingqing:
This is good for assertion but not good to introduce certain transform rules.

Consider this:

```Limit->Agg->shuffle->scan```

we can transform with globalAgg+localAgg rule against "Agg" group:

```Limit->GlobalAgg->shuffle->LocalAgg->scan```

However, the newly generated "GlobalAgg" subtree does not have the same signature as the original "Agg" subtree - for exampe: "Agg" is count(*), then "GlobalAgg" shall be sume(local_count(*)).

Currently we assert newmember memo signature must equal to existing members, so we can insert the newmember to the group list. Once we relax it, we insert into the gorup list if they have same signature, otherwise create a new group and fixing the orginal group's parent so they can reference it.

Other details:
- the original group's parent may have multiple
- if we again apply the "anti-rule" (for above example, is merge_global_local_to_agg rule), we will generate duplicate group as original group, but memo will not allow it to insert.

arzuschen:
The assertion that members within the same memo group must have the same signature does not need to be relaxed.
For the newly created global aggregate-local aggregate pair, the new global shall be in the same group as the old aggregate. This can be achieved by forcing the new one have the same signature as the original one. 
Meanwhile, the local aggregate need to be placed in a new group, this process requires identification that the node does not belong to the old aggregate group. Lifting the assertion may result in more unnecessary change to the whole exploration process.
Additionally, the property enforced nodes are added to the group after exploration. Even though the signature is different, it does not conflict with the validation step during exploration.

zhouqingqing:
closed as not needed

PR #212, opened 2020-09-02
Add exploration rule for global aggregate split

arzuschen:
Based on the assumption that aggregation cannot increase the cardinality, or that the cardinality is less or the same as before the aggregation. Therefore, it can be assumed that it is likely to be better to split the aggregate into a pair of global agg and local agg. The default of all aggregation is global aggregation, which only accept serialized data stream.
A exploration transformation rule is introduced to account for this situation. For a global aggregation that cannot be directly turned into local aggregation, it will be turned into two aggregation, a global on top of a local. The new global logic node will be placed in the same group as the original global aggregation, with its child, the new local logic node placed in a new group, and its child is the same child group as the original global aggregation.
For global aggregation, since it should be placed in the same group, the signature is forced to be the same as the original aggregate. Meanwhile, the newly created local agg node will have different signature and thus will be in its own memo group. (#143)

The aggregation functions are transformed as listed below: _(e represents some expression)_
sum(e) = sum(sum(e))
min(e) = min(min(e))
max(e) = max(max(e))
count(e) = sum(count(e))
count(*) = sum(count(*)
avg(e) = sum(sum(e)) / sum(count(e))

For groupby attribute, the local and global should use the same groupby as the original aggregate.
For having constrain, it needs to be separated, the expression involving aggregate function will be placed on the global aggregate node, while the others are place on the local aggregate node. For example, having count(a1) > 1 can only be determined after the whole aggregate process is complete, thus must be on the global node. Meanwhile, having a1 > 1 is just a local constraint that can be acted on rows, so it can be placed just on local node.

Another critical fix is that previously the aggrcore in PhysicHashAgg.Exec() is shared among threads (different copy of the phyiscal node). Now, new copy of logic_aggrFns_ is created for each thread.
<code>List<AggFunc> aggrcore = new List<AggFunc>();
logic.aggrFns_.ForEach(x => aggrcore.Add(x.Clone() as AggFunc));</code>
(line 1161 of PhysicNode.cs)

TODOs:
Stddev is not implemented in this patch, therefore query like
`select a3/2*2, sum(a3), count(a3), stddev_samp(a3) from a group by 1;`
cannot be successfully executed.
Also,
`select d1, sum(d2) from (select c1/2, sum(c1) from (select b1, count(*) as a1 from b group by b1)c(c1, c2) group by c1/2) d(d1, d2) group by d1;`
(both queries mentioned above are from TestAggregation unit test)

### DIFF
--- a/qpmodel/RulesTrans.cs
+++ b/qpmodel/RulesTrans.cs
@@ -43,6 +43,7 @@ namespace qpmodel.optimizer
         public static List<Rule> ruleset_ = new List<Rule>() {
             new JoinAssociativeRule(),
             new JoinCommutativeRule(),
+            new AggSplitRule(),
             new Join2NLJoin(),
             new Join2HashJoin(),
             new Join2MarkJoin(),
@@ -80,6 +81,8 @@ namespace qpmodel.optimizer
                 ruleset.RemoveAll(x => x is Agg2StreamAgg);
             if (!option.optimize_.enable_nljoin_)
                 ruleset.RemoveAll(x => x is Join2NLJoin);
+            if (!option.optimize_.memo_use_remoteexchange_)
+                ruleset.RemoveAll(x => x is AggSplitRule);
         }
 
         public abstract bool Appliable(CGroupMember expr);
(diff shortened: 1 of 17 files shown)


qpmodel/RulesTrans.cs
```diff
@@ -197,4 +200,125 @@ public override CGroupMember Apply(CGroupMember expr)
...
+            var localfns = new List<Expr>();
+
+            // record the processed aggregate functions
+            var derivedAggFuncDict = new Dictionary<Expr, Expr>();
+
+            foreach (var func in aggfns)
+            {
+                Expr processed = null;
+                // average is transformed into sum(sum()) / sum(count())
+                if (func is AggAvg)
```
zhouqingqing, line 269: is it possible to move these logics sitting together with corresponding aggfns?

zhouqingqing:
@pkommoju Please take a look at ResolveOrdinal related part. This change splits aggregation into one part ("gloal") on top of gather andanother part ("local") underneath gather exchange. For example, avg(a) => global = sum(bs)/sum(bc) and local = sum(a) as bs, count(a) as bc. 

Since ordinal resolution is top down, so the top node will request "avg(a)" from its child, but now the child's expression becomes sum/sum, so expr matching won't work and we use Epxr._ id field to bind the connection.

qpmodel/RulesTrans.cs
pkommoju: I am not sure I understood what you are asking about, at least, not right away. I got a general feel for the aggregate split though.

arzuschen: > is it possible to move these logics sitting together with corresponding aggfns?

That should be doable, I will move that.

PR #217, opened 2020-09-03
Change cardinality calculation for distributed plan and fix bugs

arzuschen:
The cardinality for distributed table is to be computed differently. Previously, the cardinality (card_ attributed in LogicNode) is used for computing the cost of PhysicNode. However, it can only represent serialized execution cost, and cost for distributed plan cannot be correctly represented.
In the new calculation, the card_ attribute for logic node remains the same, representing the total number of rows to be processed. A new machinecount_ attribute is introduced to logic node to identify if the node is running in parallel.

The default machinecount_ is 1, and there are three main changes involving this attribute:
1) For table scan of distributed table (distributed on a column or roundrobin), machinecount_ will be set to the number of distributions of the table. 
2) For join, the machinecount_ is the maximum of the child nodes' machinecount_. 
3) For gather, the machinecount_ will be set back to 1 to show that the output is singleton.

The Card() method for physic node now returns the average cardinality for each machine. (rounded up)
Since the cost estimation process uses the Card() method under PhysicNode, it will be using the modified cardinality, meaning the number of rows per machine. Meanwhile, logic card_ remains unchanged.

Other Bugs that are fixed in this PR:
1) False assignment of having_ during agg split: The having expression is falsely changed during application of agg split rule, which may lead to expr matching issue. It is fixed by reassigning the unchanged having_.
2) Using anydistribution strictly for distributed and not singleton: change the default distribution property to singleton to avoid ambiguity and error for join resolver enabled optimization.

### DIFF
--- a/qpmodel/LogicNode.cs
+++ b/qpmodel/LogicNode.cs
@@ -56,6 +56,7 @@ namespace qpmodel.logic
         public Expr filter_ = null;
         public List<Expr> output_ = new List<Expr>();
         public ulong card_ = CARD_INVALID;
+        public ulong machinecount_ = 1;
 
         // these fields are used to avoid recompute - be careful with stale caching
         protected List<TableRef> tableRefs_ = null;
@@ -402,7 +403,7 @@ namespace qpmodel.logic
             return card_;
         }
 
-        public ulong EstimateCard()
+        public virtual ulong EstimateCard()
         {
             return CardEstimator.DoEstimation(this);
         }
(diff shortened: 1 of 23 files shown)


zhouqingqing:
> For join, the machinecount_ is the maximum of the child nodes' machinecount_.

Is that join both sides shall have the same machinecount_?

arzuschen:
> > For join, the machinecount_ is the maximum of the child nodes' machinecount_.
> 
> Is that join both sides shall have the same machinecount_?
not necessarily, it could be that joining a singleton with a distributed, or joining a replicated with distributed, in both cases it is joining machinecount_ = 1 with machinecount_ = 10, which will result in a machinecount_ = 10 output.

Issue #215, opened 2020-09-03
Incorrect NULL argument handling in functions


pkommoju:
select substring(null, 1, 4) from a;

select repeat(null, 3) from a;

select upper(null) from a;

All fail in Bind().

PR #218, opened 2020-09-04
[bug] Incorrect NULL argument handling in functions #215

zhouruiapple:
The function expression is converted to a constant expression by normalizer when the parameter is null.
Do not need to check whether the parameter is the string type or not.

### DIFF
--- a/qpmodel/ExprFunc.cs
+++ b/qpmodel/ExprFunc.cs
@@ -222,7 +222,6 @@ namespace qpmodel.expr
         {
             base.Bind(context);
             type_ = args_()[0].type_;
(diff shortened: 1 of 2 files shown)


qpmodel/ExprFunc.cs
```diff
@@ -222,7 +222,6 @@ public override void Bind(BindContext context)
         {
             base.Bind(context);
             type_ = args_()[0].type_;
-            Debug.Assert(type_ is CharType || type_ is VarCharType);
```
zhouqingqing, line 225: how about we enhance the assertion that either type_ is char or null?

zhouruiapple: Done

Issue #35, opened 2020-01-10
[bug] bug collections - bugs before triage sit here


zhouqingqing:

zhouqingqing:
-  sql = "select * from a where a1> (select b2 from b where a1<>b1);";
 -  select a1+2 from a group by a1+3 
	- we did have a check to assert only group by exprs can show up in selection list but obvious it is broken now

zhouqingqing:
more：
-             sql = "select a1, (select count(b2) from b where b1=a1 and b2>2), (select count(b2) from b where b1=a1 and b3>3) from a;";
working if change second count(b2) => count(b3). 

- sql = "select a1, (select count(*) from b where b1=a1) from a;";

working if change count(*) from count(b1).

zhouqingqing:
Update from PR #319:

- **`select abs(-a1*2), count(*) from a group by round(a1, 10)`**: Now correctly rejected with "must appear in group by clause" error (commit 077ae09). Previously returned wrong results by using the grouping expression value for abs(-a1*2).

- **`select count() ...`**: Now correctly rejected with "count() requires an argument or count(*)" error.

Other items in this collection still need investigation (subquery order by with negation, multiple scalar subqueries with same column).

PR #222, opened 2020-09-07
fix count star related issues

pkommoju:
COUNT(*) has tableRefs now to push it to the correct side in join queries.

Bind is forgiving. Calling bind() more than once is a no-op, this change was required to deal with
the fact that group by and order by replace the column by select list column and this is already taken
into account in group/order binding but only for one level. If order by (abs(-col)) is the expression
col would be found to be bound and an assert would go off. This change avoid this problem.

### DIFF
--- a/test/UnitTest.cs
+++ b/test/UnitTest.cs
@@ -2306,6 +2306,13 @@ namespace qpmodel.unittest
             // FAILED
             sql = "select * from (select * from a join b on a1=b1) ab join (select * from c join d on c1=d1) cd on a1+b1=c1+d1"; // FIXME
             sql = "select * from (select * from a join b on a1=b1) ab join (select * from c join d on c1=d1) cd on a1+b1=c1 and a2+b2=d2;";
+
+            // COUNT(*)
+            sql = "select * from (select count(*) from a, b where a1 <> b1 and a2 <> b2) s1, (select count(*) from a, b where a1 <> b3 and a2 <> b4) s2, (select count(*) from a, b where a1 < b1 and a2 < b2) s3, (select count(*) from a, b where a1 <> b1 and a2 <> b2) s4";
+            // There is no easy way to verify the plan itself but the
+            // results should 6,8,3,6
+            result = ExecuteSQL(sql);
+            Assert.AreEqual("6,8,3,6", result[0].ToString());
         }
 
         [TestMethod]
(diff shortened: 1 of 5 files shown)


test/UnitTest.cs
```diff
@@ -1934,6 +1934,35 @@ public void TestExpr()
             TU.ExecuteSQL(sql, "0,1,2,3");
             sql = "select a1 from a where not not not a1 in (1)";
             TU.ExecuteSQL(sql, "0;2");
+
+            // From issue #35, some of the failing ones are passing now
+            // BUG: Incorrect results but no longer insists that a1 show up in group by list
+            sql = "select abs(-a1*2), count() from a group by round(a1, 10);";
+            // There are two more bugs: count() should raise error, results should be the following
```
zhouqingqing, line 1941: which # tracking these issues?

pkommoju: abs(-a1 * 2) is in issue #35. I have added it to unit tests.
I have also added:
select (select a1 from a order by -a1 limit 1), count(a1) from a group by (select a1 from a order by -a1

I tried to add
select a1, (select count(b2) from b where b1=a1 and b2>2), (select count(b2) from b where b1=a1 and b3>3) from a;

but it was returning different output than what I was asserting, no matter what. So I disabled it.

I forgot to add the following negative tests. They are returning error, which is the correct behavior.

select a1, sum(a1) from a group by a1 having sum(a2) > a3;
select * from a having sum(a2) > 1;

zhouqingqing: I have opened #223 to triage executor related issues. If you think the issue is out of the algebrizer/optimizer's range, let's put them in #233.

pkommoju: I think
select a1, (select count(b2) from b where b1=a1 and b2>2), (select count(b2) from b where b1=a1 and b3>3) from a;
and
select (select a1 from a order by -a1 limit 1), count(a1) from a group by (select a1 from a order by -a1
may or may not be executor issues.

What is noticed in "select abs(-a1*2), count(*) from a group by round(a1, 10)" is that the value fed into abs() is not (-a1 * 2), instead it is round(a1, 10). I think the same thing happens in the two statements above. So it may be group by ordinal is wrong. I will look into it again. For now, may be they can be in issue #35.

PR #225, opened 2020-09-10
Adding some of the NIST tests to excercise GROUP BY, HAVING, AGGREGATES

pkommoju:
Adding some of the NIST tests to exercise GROUP BY, HAVING, AGGREGATES and some subqueries.
There are many more relevant tests but some features are not supported
in the current code base. String comparison will be enhanced and then
some more tests will be added.
Fixed Minor typos in error/comment.

Character data comparison with character data is now supported but no implicit conversions to and from character data are supported.


test/UnitTest.cs
```diff
@@ -3141,4 +3141,769 @@ public void TestParenthesisError()
             RawParser.ParseSqlStatements(sql);
         }
     }
+
+    [TestClass]
```
zhouqingqing, line 3145: can we add a separate file for this? unittest is already too big

pkommoju: Yes, but how do I integrate into existing framework? If all of NIST stuff is moved into NistTests.cs, can they be called from the existing one by invoking RunNistTests()?

pkommoju: Moved to test/NistTests.cs. The string compare tests are in their own file, StringCompare.cs.

Issue #72, opened 2020-03-03
accouting subquery's cost into total cost


zhouqingqing:
```
PhysicOrder  (inccost=24.1, cost=0.1, rows=1) (actual rows=0)
    Output: supplier.s_name[0],supplier.s_address[1]
    Order by: supplier.s_name[0]
    -> PhysicHashJoin  (inccost=24, cost=13, rows=1) (actual rows=0)
        Output: supplier.s_name[1],supplier.s_address[2]
        Filter: supplier.s_nationkey[3]=nation.n_nationkey[0]
        -> PhysicScanTable nation (inccost=1, cost=1, rows=1) (actual rows=1)
            Output: nation.n_nationkey[0]
            Filter: nation.n_name[1]='CANADA'
        -> PhysicScanTable supplier (inccost=10, cost=10, rows=10) (actual rows=10)
            Output: supplier.s_name[1],supplier.s_address[2],supplier.s_nationkey[3]
            Filter: supplier.s_suppkey[0] in @1
            <InSubqueryExpr> cached 1
                -> PhysicFilter  (inccost=14410, cost=753, rows=753) (actual rows=47)
```
Though this does not affect the correctness since subquery  is optimized separately but better to fix for consistency.

9DemonFox:
**SQL**
```
select	(select count(O.o_orderkey) 
	 	from orders O 
	 	where O.o_custkey = 37) as OrderCount, 
	c_name,
	(select count(O2.o_custkey) 
		from orders O2
		where O2.o_custkey = 26) as CustPhone
from customer C
where exists
	(
		select *
		from nation N
		where N.n_name like '%A%'
		and exists(
			select * 
			from region R
			where N.n_regionkey = R.r_regionkey
			and R.r_regionkey<2
		)
	)
```
**physic plan**
```
order by c_name
Total cost: 3136.59, memory=4950
PhysicOrder  (inccost=3136.59, cost=766.59, rows=150, memory=4950) (actual rows=150)
    Output: c_name[0],{@1}[1],{@2}[2]
    <ScalarSubqueryExpr> cached 1
        -> PhysicHashAgg  (inccost=1541, cost=3, rows=1, memory=2) (actual rows=1)
            Output: {sum({count(o_orderkey)})}[0]
            Aggregates: sum({count(o_orderkey)}[0])
            -> PhysicGather Threads: 10 (inccost=1538, cost=10, rows=1) (actual rows=10)
[...]

PR #228, opened 2020-09-15
accouting subquery's cost into total cost  #72 

9DemonFox:


qpmodel/ClassDiagram1.cd
```diff
@@ -0,0 +1,1825 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
zhouqingqing, line 1: What is this file?

qpmodel/codegen.cs
```diff
@@ -42,7 +42,7 @@
 // Code Generation Targets:
 //  1. specialization with all possible without breaking code structure.
 //  2. can be integrated gradually - we can do one iterator by another.
-//    2.1 expr is not included
+//    2.1 expr is not included+
```
zhouqingqing, line 45: please remove this accidental change

test/UnitTest.cs
```diff
@@ -384,6 +384,67 @@ public void TestTpchDistributed()
             }
         }
 
+        [TestMethod]
```
zhouqingqing, line 387: why do we need an extra test for this fix?

qpmodel/PhysicalNode.cs
```diff
@@ -173,7 +173,21 @@ public double InclusiveCost()
                 else
                     incCost += x.InclusiveCost();
             });
-
+            //FIXED  @issue72 here should add the cost of subquery
+            if (this.logic_.filter_ != null)
```
zhouqingqing, line 177: Subquery can show up in selection, filter and some other places - so only handle filter is not enough.

qpmodel/ClassDiagram1.cd
9DemonFox: This is a UML generated by VS , I use it to understand the code,  I will use. gitignore to delete it.

qpmodel/PhysicalNode.cs
```diff
@@ -173,7 +173,21 @@ public double InclusiveCost()
...
+            {
+                Expr expr = this.logic_.filter_;
+                expr.VisitEachT<SubqueryExpr>(x =>
+                {
+                    if (x.query_.physicPlan_ != null)
```
zhouqingqing, line 182: Do you have a test case x.query_.physicPlan_ is null? That's a bug shall be fixed (you can open an issue and fix separately).

qpmodel/PhysicalNode.cs
zhouqingqing: let's first get filter case fixed and checkin other cases (selection etc) in a separate patch as it is more extensive change.

qpmodel/PhysicalNode.cs
9DemonFox: OK. And the new issue is #230

qpmodel/codegen.cs
9DemonFox: Sorry, I will check carefully next time.

test/UnitTest.cs
9DemonFox: I delete the unittest. q20 of TCP-H is more complex for debugging. So I construct a easy one for this issue.
I have renamed the TestForIssue72 to TestSubqueryInclusiveCost and move it to class Subquery in UnitTest.cs.

qpmodel/PhysicalNode.cs
9DemonFox: OK. 
And I have find this SQL with uncorrealted suquery
"select c_name,(select count(O.o_orderkey) from orders O where O.o_custkey=37) as OrderCount from customer C"
has a wrong inccost.
image

test/UnitTest.cs
zhouqingqing: I am not sure I got it - the original tpch test already have q20. Why you introduce TestSubqueryInclusiveCost() just to test q20 again?

qpmodel/PhysicalNode.cs
```diff
@@ -1,4 +1,4 @@
-﻿/*
+﻿ /*
```
zhouqingqing, line 1: extra ' '

test/UnitTest.cs
9DemonFox: The query result you quoted  in the issue description is the  result  of  q20. I test q20 for emphasis the issue of bug in filter subquery.  If you think it is unnecessary to creat a unit test  to test q20 again, i will delete it.

9DemonFox: The query result you quoted  in the issue description is the  result  of  q20. I test q20 for emphasis the issue of bug in filter subquery.  If you think it is unnecessary to creat a unit test  to test q20 again, i will delete it.

9DemonFox: The query result you quoted  in the issue description is the  result  of  q20. I test q20 for emphasis the issue of bug in filter subquery.  If you think it is unnecessary to creat a unit test  to test q20 again, i will delete it.

qpmodel/PhysicalNode.cs
9DemonFox: sorry，I will soon  amend it

test/UnitTest.cs
zhouqingqing: please remove this unittest as it is redundant.

zhouqingqing:
Thanks, merged!

9DemonFox:
Thanks for your advises. And I am not very familiar with details of git, so I have made some very low-level mistake.  Thank you again.

PR #238, opened 2020-10-14
unit test consistancy

9DemonFox:


test/UnitTest.cs
```diff
@@ -140,6 +140,23 @@ bool FindInLevel(List<PhysicNode> curlevel, string pattern, out List<PhysicNode>
             }
             return true;
         }
+
+        public static void ClearTableStatsInCatalog(List<String> tabNameList)
+        {
+            foreach (String tabName in tabNameList)
+            {
+                try
```
zhouqingqing, line 148: please remove try/catch here - I'd rather to see a more obvious panic error message instead of hide the errors.

test/UnitTest.cs
```diff
@@ -162,6 +179,39 @@ public void TestCSVReader()
             Assert.AreEqual("2,2,3,4", r[1]);
             Assert.AreEqual("5,6,7,8", r[2]);
         }
+
+        [TestMethod]
+        // it should exec after UBenchMarks
+        // the "row" in physicplan is stasticed based on the Hist or MCV, whitch is maintained in Catlog 
+        public void TestUnitTestConsistancy()
```
zhouqingqing, line 186: I don't think we need a test just to verify consistency. Instead, you can insert the "clear stats" fix at the right place?

zhouqingqing:
checkin comments please

test/UnitTest.cs
9DemonFox: got it.

qpmodel/Statis.cs
```diff
@@ -585,9 +585,17 @@ public class SysStats : SystemTable
...
-            records_.Clear();
+            var table = Catalog.systable_.Table(tabName);
+            var columns = table.columns_;
+
+            List<ColumnStat> stats = new List<ColumnStat>();
+            foreach (var v in columns)
+            {
+                var colName = v.Value.name_;
+                records_.Remove(tabName + colName);
+            }
```
9DemonFox, line 598: In case not to clear some build-in table log, only delete table records related.

Issue #197, opened 2020-08-18
Distributed TPCH incorrect results


arzuschen:
currently with redistribution and gather directly added from logic tree:
q01, q03, q05, q06, q10, q11, q12, a16, q19 are runnable
q01, q03, q06, q10, q12 have output

q10 has incorrect result due to NLJoin with distributed data stream

arzuschen:
#194 q10 incorrect result fixed.

arzuschen:
q04 has unstable output, q18 and q20 have unstable plan

9DemonFox:
#263 , they might have the same cause. The FROM clause in TPCH is not handled correctly. I am working to check it.

PR #264, opened 2020-11-05
#issue72 account subquery cost of scalarSubquery in SELECT clause to inclusive cost

9DemonFox:
meanwhile, it solves the problem that the subqueries in SELECT clause are wrongly push down


9DemonFox:
Issue #3 has raised this bug. Now filter in FROM clause can be push down. 
My method is to add the parameter **isToppest** to filter the **subqueryExpr**. And add **ExecProject** in the **physicNode** to trigger the subquery to execute.

zhouqingqing:
@pkommoju  please review ordinal related changes.

test/UnitTest.cs
```diff
@@ -453,6 +454,20 @@ public void TestUsingTpch0001Data()
             }
             List<String> tabNameList = new List<String> { "region", "orders", "part", "partsupp", "lineitem", "supplier", "nation" };
             TU.ClearTableStatsInCatalog(tabNameList);
+
+            try
+            {
+                // sql06 does not have ORDER, so the toppest physice node is physicGather
+                var badQueries = new string[] { "sql01", "sql02", "sql03", "sql04", "sql05", "sql07", "sql08" };
```
zhouqingqing, line 461: what's wrong with these bad queries?

test/regress/expect/tpch0001_select/sql01.txt
```diff
@@ -26,35 +26,9 @@ PhysicOrder  (inccost=2416.59, cost=766.59, rows=150, memory=4350) (actual rows=
...
-                    -> PhysicHashAgg  (inccost=1528, cost=28, rows=1, memory=8) (actual rows=1, loops=10)
-                        Output: {count(o_orderkey)}[0]
-                        Aggregates: count(o_orderkey[0])
-                        -> PhysicScanTable orders as o (inccost=1500, cost=1500, rows=26) (actual rows=2, loops=10)
-                            Output: o_orderkey[0]
-                            Filter: o_custkey[1]=37
-
+        Output: c_name[0]
         -> PhysicScanTable customer as c (inccost=150, cost=150, rows=150) (actual rows=15, loops=10)
-            Output: c_name[1],@1
```
zhouqingqing, line 44: This plan change does not look right: @1 shall go with customer table since it is part of selection list?

qpmodel/stmt.cs
```diff
@@ -782,13 +783,15 @@ Expr RepalceSuquerySelection(Expr e)
             return newselection;
         }
 
+
         internal void ResolveOrdinals()
         {
+            bool isToppestNode = true;
```
pkommoju, line 787: Won't this be true for every recursive call to resolve sub-query ordinals? Is that the intent? On the otherhand, if only the outer most query should have this to set to true, then context->parent_ == null would work, doesn't it?

9DemonFox:
image
@zhouqingqing , I solve the the subquey cost in two ways.
PR#240 will unnested scalar whether it is correlated or not. The Hyper used this idea. 
image

zhouqingqing:
> PR#240 will unnested scalar whether it is correlated or not

Why unnest non-correlated scalar is a better idea? Is it execution less efficient?

9DemonFox (commented):
> Why unnest non-correlated scalar is a better idea? Is it execution less efficient?
To my understanding, they have the same time complexity. 
```
Unnest: 1 Join N = O(N)
Cached: for r in R:
                 r = r,1
```
Actully efficient may need to be verified on product enviroment...

9DemonFox:
@zhouqingqing 
> Why unnest non-correlated scalar is a better idea? Is it execution less efficient?
To my understanding, they have the same time complexity. 
```
Unnest: 1 Join N = O(N)
Cached: for r in R:
                 r = r,1
```
Actully efficient may need to be verified on product enviroment...

9DemonFox:
#issue263 Fixed, and some in #197 also fixed.

PR #270, opened 2020-11-13
enable optimize_.enable_subquery_unnest_ = false

9DemonFox:


9DemonFox:
The SQL below will trigger this bug
```
  sql = "select a2/2, count(*) from (select a2 from a where exists (select * from a b where b.a3>=a.a1+b.a1+1) or a2>2) b group by a2/2;";
                TU.ExecuteSQL(sql, "0,1;1,2", out phyplan, option);
```
because it will remove from and not let it be added to subQueries, i.e., **option.optimize_.remove_from_**  and
**option.optimize_.enable_subquery_unnest_** are not indepwndent
```
                    var plan = qref.query_.CreatePlan();
**                    if (qref is FromQueryRef && queryOpt_.optimize_.remove_from_)
**                        from = plan;
                    else
                    {
                        string alias = null;
                        if (qref is CTEQueryRef cq)
                            alias = cq.alias_;
                        else if (qref is FromQueryRef fq)
                            alias = fq.alias_;
                        var key = new NamedQuery(qref.query_, alias);
                        from = new LogicFromQuery(qref, plan);
                        subQueries_.Add(key);

                        // if from CTE, then it could be duplicates
                        if (!fromQueries_.ContainsKey(key))
                            fromQueries_.Add(key, from as LogicFromQuery);
                    }
```

qpmodel/Plan.cs
```diff
@@ -344,7 +344,7 @@ LogicNode transformOneFrom(TableRef tab)
                         break;
                     case QueryRef qref:
                         var plan = qref.query_.CreatePlan();
-                        if (qref is FromQueryRef && queryOpt_.optimize_.remove_from_)
+                        if (qref is FromQueryRef && queryOpt_.optimize_.remove_from_&&queryOpt_.optimize_.enable_subquery_unnest_)
```
9DemonFox, line 347: enable_subquery_unnest_ and remove_from are not independent, without this condition, the from query will not add to **subqueries**. And the **subquris** will not have physic plan. see CopyOutOptimalPlan
``` 
            foreach (var v in subqueries)
                CopyOutOptimalPlan(v.query_, !select.SubqueryIsWithMainQuery(v));
```

zhouqingqing: We may want to decouple them: 
- remove_from is used to collapse multiple FROM subquery, which is not correlated. Say "select * from (select * from a) b;" becomes "select * from a". You shall see no more LogicFromQuery in plan. 
- subquery_unnest is to decorrelate correlated subqueries.

So logically they are independent.

9DemonFox: OK, I'll check it in detail latter. The PR is just a workround.

PR #274, opened 2020-11-18
Andb parser int

pkommoju:
ANDB parser integration into native code.

Please read qpmodel/docs/using_andb_parser.txt before attempting to build it on Windows. I have not built native code on Linux, therefore I am not sure how the CMake system may have to be adapted to produce the effects of the script the document refers to.
Also to note is the fact that server/experimental/src/common has a public domain getopt header and C file to implement the option processing required for andb.exe (the command line driver for the native code). getopt.cpp needs to be excluded from SRC_LIST in server/experimental/src/common/CMakeLists.txt when running on Linux.


server/experimental/src/common/getopt.cpp
```diff
@@ -0,0 +1,188 @@
+/*****************************************************************************
```
zhouqingqing, line 1: gflags lib already have function like/better than this - let's try to use gflags.

pkommoju: I will look into it.

server/hyrise-includes/ColumnType.h
```diff
@@ -0,0 +1,18 @@
+#ifndef __COLUMNTYPE_H_
```
zhouqingqing, line 1: let's try to remove other dependencies on hyrise code except the parser syntax file part

pkommoju: I will remove them.

pkommoju: done

server/experimental/src/common/getopt.cpp
pkommoju: I looked into it and think it is too heavy for the temporary purpose of command line driver, which only needs to be told if the input is from a file or should it become interactive. I can get rid of getopt and let andb.exe become interactive if argv[1] is NULL, or treat argv[1] as input file, which will avoid the issue of command line argument processing.

PR #275, opened 2020-11-18
#issue83 tpcds

9DemonFox:
save tpcds result into files. This PR can be merged firstly to see how the output files change.


test/UnitTest.cs
```diff
@@ -385,7 +385,7 @@ public void TestJobench()
         public void TestBenchmarks()
         {
             TestTpcdsPlanOnly();
-            TestTpcdsWithData();
+            //TestTpcdsWithData();
```
zhouqingqing, line 388: is this accidentally disabled?

zhouqingqing:
is this pr still needed?

9DemonFox:
> is this pr still needed?
nope, as the PR277 has override it

PR #278, opened 2020-11-25
Extract tpcds tests

9DemonFox:
generate tpcds output files for comparation.


9DemonFox:
original cpcds test turn on all opt..

PR #283, opened 2020-12-01
Andb parser int

pkommoju:
Binding selections, which at the present are columns and simple expressions
(+, -, *, =, <=) is done.

Binding WHERE is done (Untested).

Plan generation is being integrated:
   a) Create Plan (Only LogicScan),
   3) Create Physical Plan
   c) Open Physical Plan

   are now connected to andb driver code.

Execution is still not connected.

andb.exe defaults to interactive mode, -f file turns it off.

Code formatting switched from clang to "regular" (for VIM navigation)
and to avoid having to deal with all sorts of extraneous punctuation put in
by "clang" auto-formatting.


zhouqingqing:
please add your format file change, otherwise the format adjustment will be fragile.

PR #290, opened 2020-12-17
Andb parser int

pkommoju:
PhysicScan is begining to work with base tables.
It needs to go through a Project operator, without a project, it now works like select * from table.
Test table d did not get populated, fix it ASAP.


zhouqingqing:
shall we collapse projection into scan?

pkommoju:
I think we could merge project into scan.

PR #307, opened 2021-01-29
CTE optimization based on CTE producer/consumer model

9DemonFox:
1. inline cte only use one time
2. cost based cte
3. predict pushdown in cte is handled now

still need to perfect:
1. FROM can handle alias tabel name, so I don't  emit eliminate FROM under CTEProducer and CTESelect.
2. CTEAnchor Node and CTESelect Node  do nothing and they should be eliminated too.
3. subquery in CTE is not well handled due to some thing in Bind. so q58 and q83 in tpcds can't run well.


test/regress/expect/tpch0001/q22.txt
```diff
@@ -35,20 +35,20 @@ group by
...
         Group by: cntrycode[0]
-        -> PhysicFromQuery <custsale> (inccost=231730, cost=2250, rows=2250) (actual rows=9)
+        -> PhysicFromQuery <custsale> (inccost=231440, cost=2250, rows=2250) (actual rows=9)
             Output: cntrycode[0],c_acctbal[1]
-            -> PhysicFilter  (inccost=229480, cost=2250, rows=2250) (actual rows=9)
+            -> PhysicFilter  (inccost=229190, cost=2250, rows=2250) (actual rows=9)
                 Output: {substring(c_phone,1,2)}[0],c_acctbal[1]
                 Filter: {#marker@2}[2]
-                -> PhysicMarkJoin Left (inccost=227230, cost=225000, rows=2250) (actual rows=17)
+                -> PhysicMarkJoin Left (inccost=226940, cost=225000, rows=2250) (actual rows=17)
```
zhouqingqing, line 51: how come markjoin cost changed?

tpch/q15.sql
```diff
@@ -11,8 +11,8 @@ with revenue0 as
 		and l_shipdate < date '1996-01-01' + interval '3' month
 	group by
 		l_suppkey)
-
-
+		
+
```
zhouqingqing, line 15: revert changes here

qpmodel/DataType.cs
```diff
@@ -390,7 +399,6 @@ public override List<Expr> AllColumnsRefs(bool refresh = false)
                 // make a coopy of selection list and replace their tabref as this
                 var r = new List<Expr>();
 
-                Debug.Assert(query_.bounded_);
```
zhouqingqing, line 393: before Bind(), AllColumnRefs() does not return right result?

qpmodel/PhysicalNode.cs
```diff
@@ -166,15 +166,21 @@ public double InclusiveCost()
         {
             var incCost = 0.0;
             incCost += Cost();
+
             children_.ForEach(x =>
             {
                 if (x is PhysicMemoRef xp)
                     incCost += xp.Group().nullPropertyMinIncCost;
                 else
+                {
```
zhouqingqing, line 175: is this the reason that many plans have a small cost change?

qpmodel/PhysicalNode.cs
```diff
@@ -1529,61 +1543,37 @@ protected override double EstimateCost()
...
         public PhysicFromQuery(LogicFromQuery logic, PhysicNode l) : base(logic) => children_.Add(l);
 
-        public bool IsCteConsumer(out CTEQueryRef qref) => (logic_ as LogicFromQuery).IsCteConsumer(out qref);
+        public PhysicFromQuery(LogicCteConsumer logic, PhysicNode l) : base(logic) => children_.Add(l);
 
         public override void Open(ExecContext context)
         {
             base.Open(context);
-            var logic = logic_ as LogicFromQuery;
-            if (logic.IsCteConsumer(out CTEQueryRef qref))
```
zhouqingqing, line 1544: So with a sequence plan, a cteProducer generates a query result, later cteConsumers can reuse the result - it is used to be in cteCache_, where is that logic now?

qpmodel/RulesImpl.cs
```diff
@@ -190,18 +190,41 @@ public override CGroupMember Apply(CGroupMember expr)
...
+            return log != null;
+        }
+
+        public override CGroupMember Apply(CGroupMember expr)
+        {
+            var log = expr.logic_ as LogicSelectCte;
+
+            // FIXME we need to delete FROM here
+            PhysicNode pfq = new PhysicSelectCte(log, new PhysicMemoRef(log.child_()));
+
```
zhouqingqing, line 207: remove extra spaces in this file

zhouqingqing (commented):
Can you add a description how it is worked?

qpmodel/DataType.cs
9DemonFox: unexpect change reverted.

qpmodel/PhysicalNode.cs
9DemonFox: cte name will be registerd in CTEProducer when Bind CTEProducer.  and cte consumer can read from the context.

9DemonFox (commented):
@zhouqingqing some unexpect changes have been reverted and marked as `outdate`

9DemonFox:
# 1. InitCteInfo

image


every SQLstatment has a member CteInfo, and each has a CteInfoEntries. For convenience, all of the members are declared as PUBLIC.

## 1.1 See cte as "select * from cteQuery", so it's a **SelectStmt**. and we can bind and create plan for it.

when meet `with cte as (cteQuery)`,  create a plan for it , see `CteSelectStmt `'s constructor.

```
    public class CteSelectStmt : SelectStmt
    {
        private CteExpr originCte_;
        public int CteId() => originCte_.cteId_;
        // we have to constract a CteSelection
        // select * from cte
        public CteSelectStmt(CteExpr originCte, CTEQueryRef from, List<CteExpr> previousCtes) : base(
                    new List<Expr>() { new SelStar(originCte.cteName_) },
                    new List<TableRef>() { from },
                    null, null, null,
                    previousCtes, null, null, null, $"select * from {originCte.cteName_}")
        {
            originCte_ = originCte;
        }
    }
```

## 1.2 add previous cte to current cte's context 

in **InitCteInfo**, bind and create plan for each CTE, and add previous CTEs into current cte’s context and **ctes_**, so if current cte refer to previous ctes, it can find them.

```
List<CteExpr> previousCtes = cteInfo_.GetAllCteExprs();

// prevoious ctes have smaller cteId
Debug.Assert(previousCtes.All(x => x.cteId_ < curCte.cteId_));

SelectStmt cieStmt = new CteSelectStmt(curCte, from, previousCtes);
```

## 1.3 add cte to subquery's context so subqueries can find ctes

# 2. how to delete unused CTE

## 2.1 Example

It seems easy to do so, if a `cte` is not referred, it should be delete, but the paper put forward a extreme stupid situation. see:

**sql1**:

```
"with cte0 as (select * from a),cte1 as (select * from cte0) select * from b where b.b1 = 2";
```

cte0 has been referred by cte1, but cte1 is never used. here we use a stack to remove cte0.

```
TU.ExecuteSQL(sql, "2,3,4,5", out phyplan, option);
var answer = @"PhysicScanTable b (actual rows=1)
                Output: b.b1[0],b.b2[1],b.b3[2],b.b4[3]
                Filter: b.b1[0]=2";
TU.PlanAssertEqual(answer, phyplan);
```

## 2.2 Algorithms

```
Algorithm findUnusedCTE
INPUT: cteStack
OUTPUT: cte will be marked used or not.
step1 	findCteConsumer(rootPlan,&cteStack)
step2 
        for subq in subqueres:
            findCteConsumer(subq,&cteStack)
step3   while not cteStack.empty():
            cteid = cteStack.pop()
            findCteConsumer(GetCteInfoEntryByCteId(cteid).plan_,&cteStack)
```

`findCteConsumer` will find `CteConsumer` from `LogicNode` recursively, push and mark `cteInfoEntry` as used.

For instance, in **sql1**, step1 will not push `cte1` into `cteStack`,  step2 will not either. so the cte1 will not be marked used. so the cte0 will not be marked used too. 

## 2.3 Confirm the Algorithm and Explan the example

For **sql1**, cte1 will not be marked used, so it will not push into the cteStack, so cte1 is not mark used. so they are all deleted.

# 3. Logic Plan 

image


## 3.1 LogicCteAnchor

### 3.1.1 where Cte Anchor show up. 

the `LogicCteAnchor` will be add at the root of a plan, such as (b) .

### 3.1.2 where is the code

the realize code is bellow in `cteToAnchor`

```
            for (int i = 0; i < cteIsUsed.Count; i++)
            {
                var lca = new LogicCteAnchor(cteInfo_.GetCteInfoEntryByCteId(cteIsUsed[i].cteId_));
                lca.children_.Add(root);
                root = lca;
            }
            return root;
```

## 3.2 LogicCteProducer

### 3.1.1 where Cte Producer show up. 

the plan produced for `select * fom (cteQuey)` is the plan for `CTEProducer`. 

the `LogicCteProdicer` will be produced through `RuleTrans`

### 3.1.2 where is the code

the child of the `LogicCteProducer` is the `select * from cteQuery` , see section 1.1.

```
        public LogicCteProducer(CteInfoEntry cteInfoEntry)
        {
            cteInfoEntry_ = cteInfoEntry;
            children_.Add(cteInfoEntry.plan_);
            // CTEProducer has only one child
            Debug.Assert(children_.Count() == 1);
        }
```

## 3.3 LogicCteConsumer

`cteQueryRef` will be transform as `LogicCteConsumer`. 

## 3.4 Logic Sequence

### 3.4.1 where it show up

`RuleTrans` `CteAnchor2CteProd` will  produce a `LogicSequenceNode`, and it has the same `MemoSignature`  with `CTEAnchor` for it.

# 4. Cost-Based Optimized CTE

## 4.1 how to produce a memo as this figure below

image


the `LogicCteProducer` will be add through `RulesTrans`. when we meet a `LogicCTEAnchor`, it will producer a `LogicSequence ` Node, and the `LogicSequence` will add `CTEProducer` Node as its child.

when meet `CTEConsumer`, it will producer `LogicCTESelect` and has the same `Soginature` with `CTEConsumer`. it will also create nodes from `cteplan`. To avoid Group7 produce  the same `Signature`  with Group6, the Signature have to consider  if it is  belong to a child of cte consumer,  `cteId` and `cteRefId`. 

| group  | is child of cte consumer | cteid | cteRefId | Signature |
| ------ | ------------------------ | ----- | -------- | --------- |
| Group6 | false                    | 1     | -        |           |
| Group7 | true                     | 1     | 1        |           |
| Group8 | true                     | 1     | 2        |           |

by this , `cteid = 2` and `cteRefid = 1` will has the same signature with `cteid = 1` and `cteRefid=2` will has the same Signature. To avoid this ,  see `cteRefId` as string. 

| group  | is child of cte consumer | cteid | cteRefId | Signature |
| ------ | ------------------------ | ----- | -------- | --------- |
| Group6 | false                    | 1     | -        |           |
| Group7 | true                     | 1     | `ref1`   |           |
| Group8 | true                     | 1     | `ref2`   |           |

## 4.2  select plan

image
Figure 4.1

### 4.2.1 how to avoid  plan(a)

A `LogicCteSequence`  which is a binary operator will bind to a `cte`, so it has its `cteid`  and will generate requirement `(p 0)` for left child and `(c 0) `for right child. And a `LogicCteAnchor` will generate no requirement for its child. 

`requirement for CTE ` will always pass to children and never disappear.  The Group Member of a Group will satisfied the requirement only when it's `cteSpec` is the sub set of the `CteReq` 

a `LogicCteAnchor` will generate no requirement for its child, when meet `consumer0`, `consumer0` will seen as [(c 0)], it is not the subset of []. So the plan will not pass.

image


### 4.2.2  c and d are valid

In fact, c and d should be avoid. But now we don't support transform property from left to right child, they are not avoid. 

But they will be optimized out by cost.

cost-based select will not choose plan like (c) and (d)

image
image

PR #321, opened 2026-04-14
Fix OR handling in inToMarkJoin and scalarToSingleJoin (#271)

### LINKED ISSUE #271: OR should be handled in scalarToSingleJoin and inToMarkJoin
we only handle it in existToMarkJoin

zhouqingqing:
## Summary

Fixes #271 — OR expressions were only handled in `existsToMarkJoin` but not in `inToMarkJoin` or `scalarToSingleJoin`. When a correlated IN/scalar subquery appeared inside an OR (e.g., `a1 IN (select ...) OR a2 > 1`), decorrelation did not preserve OR semantics.

- Apply the same OR-aware filter extraction from `existsToMarkJoin` to `extractCurINExprFromNodeAFilter` and `djoinOnRightFilter`
- Detect OR conjuncts with extra subqueries via `hasAnyExtraSubqueryExprInOR`
- Set `canReplace` flag so the caller advances `innerNode` correctly
- Pass `canReplace` ref from `oneSubqueryToJoin` to all three methods

**Note:** The nested case (OR *inside* a scalar subquery's filter alongside another correlated subquery — the original #271 example) remains unsupported during unnesting and falls back to nested-loop execution.

## Test plan
- [x] New tests: correlated IN in OR, NOT IN in OR, two scalar subqueries in OR
- [x] All 72 existing tests pass
- [x] TPC-H / TPC-DS benchmarks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zhouqingqing:
Superseded by new PR with clean single commit on dedicated branch.

